PR #20, opened 2019-09-24
Reducers

totokaka:
Adds tooling for reducers

The function `reducer` is a small helper for typing reducers and putting them in a tuple-like array with the action type they should reduce.

The function `combineReducers` takes multiple reducer entries and returns a streaming operator that uses scan to reduce incoming actions to a state by using the reducers.

TypeScript was a bit hard to work with on `reducer`; the conditional types didn't quite do the trick, so I needed manual overloads, and duplicate the documentation. Unfortunate, but not a big problem IMO.


andreajl:
To have reducers support to not use the payload for action with a payload.

`export const fetchUser = actionCreator<string>('[user] fetch');`
`const fetchUserReducer = reducer(fetchUser, (prev: State) => ({ ...prev, loading: true })`

This gives the type error: `Argument of type 'ActionCreatorWithPayload<string>' is not assignable to parameter of type 'ActionCreatorWithoutPayload'.ts(2345)`

totokaka:
This should now be ready for next round of discussion / review.
Namespaces are not an issue for reducers, but state streams, so we can deal with that later.

The biggest issue I see is reducers that should handle multiple actions. We have quite a lot of those in ardoq-front.
I can see two approaches to this, but I don't like any of them
1. Multiple actions by default. `ReducerEntry` would become `[ActionCreator[], Reducer]`, `reducer` would be `reducer([actionA, actionB], (state, payload) => ...)` or `reducer((state, payload) => ..., actionA, actionB)`
2. Split approach with overloads: `ReducerEntry` would be `[ActionCreator, Reducer] | [ActionCreator[], Reducer]` and `reducer` would support `reducer(actionA, (state, payload) => ...)` and `reducer([actionA, actionB], (state, payload) => ...)`, or even been variadic like in 1

I don't like approach 1 because we should encurage single-action reducers, and I don't like 2 because it gives us unecessary complexity with overloads and union types.

(Impossible to keep line coverage at 100%, btw)

totokaka:
@andreajl jup, that works :)

PR #25, opened 2019-11-01
Make actions readonly and protected

### LINKED ISSUE #21: Protect properties
- Actions should be immutable: `readonly`
- Action creators should have their `type` field properly immutable, `readonly` and defined with `Object.defineProperty`

totokaka:
This recursively marks all created actions (and everything in their
payloads) as `readonly`, and protects them with the use of
`Object.defineProperty` with `writable: false`.

There is a performance aspect to doing a deep copy and protecting the
properties of all created action objects, but it is measured in
microseconds. Deep copy and protecting is about [half as fast][1] as
protecting in place.

Fixes #21 

[1]: https://jsperf.com/deep-copy-and-protect/130


totokaka:
As a test I tried to run ardoq-front against this branch, and that revealed some issues.

Typescript complains in many places that it cannot ensure that the functions a payload is passed into will treat it as immutable. It will be a big task to fix the typings for this.

I did try to just ignore the typescript errors and run the app, and I could not find any places where the runtime check fails. This means the typescript errors are perhaps only typescript errors. However we have only adopted the `actionCreator` in very modern parts of the code base, so these results are not representative for the code base as a whole.

We have to consider how having deeply protected actions will affect the speed at which we can adopt rxbeach. Using temporary type casting utils like `FIXME_mutable(payload)` could be a work around, but we, or our users, might encounter runtime errors.

@chriskr has suggested instead starting with doing shallow protection of only the fields on the action itself (`payload`, `meta.namespace` and `type`).

chriskr:
What is the status here? I still think we shouldn't do it, or more properly I think we have not enough resources to do it properly. Only protecting the payloads without introducing immutable state doesn't make too much sense, only adds to the inconsistency of our code base.

totokaka:
> What is the status here? I still think we shouldn't do it, or more properly I think we have not enough resources to do it properly. Only protecting the payloads without introducing immutable state doesn't make too much sense, only adds to the inconsistency of our code base.

I agree we don't have the resources to go through with this change in front. My plan was just to leave this issue open with the `blocked-by-ardoq` tag.

chriskr:
Ok, but I remove myself as reviewer so I no longer get reminders.

Issue #23, opened 2019-10-02
Evaluate test runner/framework options 


anton164:
Currently, we've started using Mocha for running tests in this repo. However, as @holographic-principle suggested it would be interesting to consider other test frameworks, such as AVA, Jest, etc. Not because it matters too much, but it's a nice exercise and if we find a good approach we could potentially later apply the same testing approach to `ardoq-front` :)

Currently in `ardoq-front` we use Karma and Jest to run tests. Karma runs unit tests, while Jest runs our browser tests.

I would guess that one test runner could handle both, but which one is the best? 😄 

_One test framework to rule them all? 🚀_

The needs for this library is probably mainly running unit tests and support for marble testing using `rxjs-marbles`. As we add React views, it would be nice to have some tests for those as well (i.e. snapshot testing?).

In addition to those, `ardoq-front` needs a way to run browser tests using `puppeteer`, but we should be careful about imposing `ardoq-front`'s needs in this repo where it doesn't make sense :)

totokaka:
@anton164 We can close this, right?

PR #30, opened 2019-11-10
Switch to ava test framework

holographic-principle:
This is related to #23 . I integrated Ava into the project, and converted all existing tests to it. The new tests use the `*.spec.ts` extension. I also kept the old tests for now.

I also set up coverage reporting using `nyc`.

This PR (for now) doesn't cover [browser testing](https://github.com/avajs/ava/blob/master/docs/recipes/puppeteer.md) or [React testing](https://github.com/avajs/ava/blob/master/docs/recipes/react.md), but they are both supported by Ava.

You can run the tests using `yarn test-ava`. Don't forget to run `yarn install` beforehand.

For me, it takes about 10.5 seconds with the coverage report, and around 8.5 seconds if I just run `ava`. From some initial research, it seems that the slowness could be coming from a sub-optimal globbing pattern used in the configuration, or just from the default behavior of ava to fork a new process for each test file. It could also have something to do with how the tests are written, as well as the typescript compilation part.

This requires some further testing.

### DIFF
--- a/examples/actions.spec.ts
+++ b/examples/actions.spec.ts
@@ -1,0 +1,52 @@
+import test from 'ava';
+import { of } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { actionCreator } from 'rxbeach';
+import { extractPayload, ofType } from 'rxbeach/operators';
+
+const voidAction = actionCreator('[test] void action');
+const primitiveAction = actionCreator<number>('[test] primitive action');
+type Payload = { foo: number };
+const payloadAction = actionCreator<Payload>('[test] payload action');
+
+test('creates action objects', t => {
+  const a = voidAction();
+  const b = primitiveAction(321);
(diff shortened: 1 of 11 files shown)


examples/actions.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  t.is((a as any).payload, undefined);
+  t.is(b.payload, 321);
+  t.deepEqual(c.payload, { foo: 123 });
+});
+
+test('can filter actions', async t => {
+  const a = primitiveAction(12);
+  const r = await of(voidAction(), a, payloadAction({ foo: 123 }))
+    .pipe(ofType(primitiveAction))
+    .toPromise();
```
anton164: I was a bit late to the party here, but I still think we should be using marble tests for testing streams.

Converting streams to a promise is an generally considered an anti-pattern and I think it makes sense to avoid doing that in tests as well.

PR #40, opened 2019-11-25
Add void and payload overloads for actionCreator

### LINKED ISSUE #34: Look into typing of `actionCreator`
`actionCreator` currently has problems if the payload type is
- `boolean` - which is treated as `true | false`
- An enum - which is also done as a disjunction of all the enum types
- A disjunct type

totokaka:
These overloads allows TypeScript to properly infer Payloads for union payload
types.
Fixes #34.

### DIFF
--- a/src/actionCreator.tspec.ts
+++ b/src/actionCreator.tspec.ts
@@ -1,0 +1,30 @@
+import { actionCreator } from 'rxbeach';
+
+const booleanAction = actionCreator<boolean>('boolean');
+booleanAction(true);
+booleanAction(false);
+
+enum Enum {
+  A,
+  B,
+  C,
+}
+const enumAction = actionCreator<Enum>('enum');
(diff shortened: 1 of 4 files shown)


src/actionCreator.tspec.ts
```diff
@@ -0,0 +1,30 @@
...
+enumAction(e);
+
+const unionAction = actionCreator<{} | number>('empty or number');
+unionAction({});
+unionAction(1);
+unionAction(2);
+
+type NumberGenerator = () => number;
+type BoolFunc = (a: boolean) => boolean;
+const unionFuncAction = actionCreator<NumberGenerator | BoolFunc>('functinos');
```
andreajl: functinos! 🔥

totokaka:
> Looks good, but do we need this?

Yes. Without this, we can't use primitive booleans or enums as payload types.

PR #61, opened 2019-12-19
Add carry operator

andreajl:
An operator to carry the initial payload through a new operator and returns it combined as an array with results from the operator.

```ts
action$.pipe(
  extractPayload(),
  carry(map(payload => payload.foo))
  tap(([payload, foo]) => {
    // `payload.foo === foo` equals `true`
  })
)
```
Big help from Tobias 🐧

### DIFF
--- a/src/operators/index.ts
+++ b/src/operators/index.ts
@@ -2,5 +2,6 @@ export {
   ofType,
   extractPayload,
   withNamespace,
+  carry,
 } from 'rxbeach/operators/operators';
 export { mergeOperators } from 'rxbeach/operators/mergeOperators';
--- a/src/operators/operators.spec.ts
+++ b/src/operators/operators.spec.ts
@@ -5,7 +5,12 @@ import {
   actionCreator,
   namespaceActionCreator,
 } from 'rxbeach';
-import { extractPayload, withNamespace, ofType } from 'rxbeach/operators';
+import {
+  extractPayload,
+  withNamespace,
+  ofType,
+  carry,
+} from 'rxbeach/operators';
 import { mockAction } from 'rxbeach/internal/testUtils';
 import { map } from 'rxjs/operators';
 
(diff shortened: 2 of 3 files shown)


holographic-principle:
I don't quite understand...

Is `foo` always the same value after `carry` is applied? Shouldn't this be achieved through partial application rather than cloning the stream and merging it into itself (isn't that what happens by using `withLatestFrom`?).

andreajl:
> I don't quite understand...
> 
> Is `foo` always the same value after `carry` is applied? Shouldn't this be achieved through partial application rather than cloning the stream and merging it into itself (isn't that what happens by using `withLatestFrom`?).

This is maybe a dumb misleading example. But the idea is to avoid nesting the chain, or having ugly  return conditions inline. See https://github.com/ardoq/ardoq-backoffice/blob/672ff79bdee450749d138b0106b0df8c5fd5a2bd/src/app/modules/user/routine.ts#L60 for an usage. I just want to bring the payload through an operator, which probably will return something totally different from the initial payload.

holographic-principle:
Ok, so it's basically side-effecting state into the stream somewhere along the way.

andreajl:
> Ok, so it's basically side-effecting state into the stream somewhere along the way.

Yes, that's a good way to put it.

totokaka:
> Ok, so it's basically side-effecting state into the stream somewhere along the way.

The operator is stateful, yes, but there should be no side effects going on.
I thinks it's very good that you are taking a critical angle to this, btw.

holographic-principle:
How are we doing this now, and are there any alternatives that you've come across while considering this solution?

andreajl:
> How are we doing this now, and are there any alternatives that you've come across while considering this solution?

1. You can do it inline like this in some cases: https://github.com/ardoq/ardoq-front/blob/30afd9a777ecb1a0579bd4a2f23874b8d60faf22/src/js/ardoq/search/Gremlin/routines.ts#L50.
2. And this with nesting, although quite old code: https://github.com/ardoq/ardoq-front/blob/30afd9a777ecb1a0579bd4a2f23874b8d60faf22/src/js/ardoq/performanceDashboard/dashboardEpics.js#L72

holographic-principle:
If you think in terms of reducers -- pure functions that take previousState and return nextState (of the same structure), it seems to me that if you were using a ReducerMap here, you would basically be forced to include `foo` in your stream's state (even if it's empty until a certain point). And I think you should just do the same thing here, even though the inline style allows you all sorts of acrobatics.

PR #63, opened 2019-12-20
Stream dependency analysis

totokaka:
This PR adds the tools we need to analyze how streams depend on each other.

The building blocks of this is as follows:
- `Marker` - A simple object which says something about how a stream is built up. The markers actually represent a stream as a tree in itself. They tell us:
    - The name of the stream
    - The sources of the stream (streams which updates leads to this stream updating)
    - The dependencies of the stream (streams which we retrieve data from)
- `MarkerOperator` - A no-op operator that places a marker on a stream
- `findMarker` - A function which recursively looks at the operators and sources for a stream to find the closest marker

Have a look at the test for `derivedStream` to see how the structure might look.

### DIFF
--- a/src/internal/CombinationOperator.ts
+++ b/src/internal/CombinationOperator.ts
@@ -1,0 +1,53 @@
+import { Observable, OperatorFunction } from 'rxjs';
+
+export type CombinationOperator = {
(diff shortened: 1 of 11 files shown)


totokaka:
I have merged this, but please do review it!

holographic-principle:
:scream: ;)

PR #80, opened 2020-01-27
WIP transition

holographic-principle:


src/reducer.ts
```diff
@@ -131,20 +131,20 @@ export const combineReducers = <State>(
   reducers: ReducerEntry<State, any>[]
 ): OperatorFunction<UnknownAction, State> => {
   const reducersByActionType = new Map(
-    reducers.flatMap(([actions, reducer]) =>
-      actions.map(action => [action.type, reducer])
+    reducers.flatMap(([actions, reducerEntry]) =>
```
anton164: IMO `reducer` is a better name here

src/internal/markers.spec.ts
```diff
@@ -119,7 +118,6 @@ test('markMerge includes sources', t => {
 });
 
 test('markZip includes sources', t => {
-  const dependency$ = source$.pipe(markName('dependency'));
```
anton164: Why did you change this? Seems out of scope for this PR

.eslintrc.js
```diff
@@ -1,19 +1,8 @@
...
-    ecmaVersion: 2018,
-    ecmaFeatures: {
-      jsx: true,
-    },
-  },
-  settings: {
-    'import/ignore': ['node_modules'],
+  extends: '../shared-configs/eslint-lib',
+  rules: {
+    'no-restricted-imports': ['error', { patterns: ['./*', '../*'] }],
```
holographic-principle: `rxbeach` used a fairly recent copied-over config from `ardoq-front`. Here, I just kept the rules that existed in addition to the standard ones. This one I guess is to force "clean" imports using `rxbeach/`.

.eslintrc.js
```diff
@@ -1,19 +1,8 @@
...
+    'func-style': ['error', 'expression'],
```
holographic-principle: This one is for `const myFunc = ...` instead of `function myFunc()...`. Just kept it because it was here already (in addition to the base config that was extracted).

.eslintrc.js
```diff
@@ -22,81 +11,11 @@ module.exports = {
         mocha: true,
       },
       rules: {
+        '@typescript-eslint/camelcase': ['off'],
```
holographic-principle: This is an override for `tspec` test files. @totokaka prefers non-camelcase for added readability in certain types of tests.

examples/namespacingActionDispatchers.spec.ts
```diff
@@ -20,17 +20,17 @@ test.before(async function() {
   const dispatchA = namespaceActionDispatcher(namespaceA, dispatchAction);
   const dispatchB = namespaceActionDispatcher(namespaceB, dispatchAction);
 
-  const a_p = action$.pipe(withNamespace(namespaceA)).toPromise();
-  const b_p = action$.pipe(withNamespace(namespaceB)).toPromise();
-  const sum_p = action$
+  const aP = action$.pipe(withNamespace(namespaceA)).toPromise();
```
holographic-principle: This is a fix for some non-camelcase variables in spec files

src/internal/markers.spec.ts
```diff
@@ -109,7 +109,6 @@ test('markWithLatestFrom includes source and dependencies', t => {
 });
 
 test('markMerge includes sources', t => {
-  const dependency$ = source$.pipe(markName('dependency'));
```
holographic-principle: Shadowing.

src/internal/markers.ts
```diff
@@ -176,9 +176,8 @@ export const findMarker = (observable$: Observable<unknown>): Marker | null => {
     return observable$.marker;
   } else if (observable$.source instanceof Observable) {
     return findMarker(observable$.source);
-  } else {
-    return null;
   }
+  return null;
```
holographic-principle: Redundant `else` after a `return` in `if` (`no-else-return`). I personally think we should use this rule.
```
if (thing) {
  return 1;
} // no else here
return 2;
```

src/reducer.ts
```diff
@@ -106,12 +106,12 @@ type ReducerFn = {
   ): ReducerEntry<State, VoidPayload>;
 };
 
-export const reducer: ReducerFn = <State>(
+export const reducer: ReducerCreator = <State>(
   actionCreator: UnknownActionCreator | UnknownActionCreator[],
-  reducer: Reducer<State, any>
```
holographic-principle: Shadowing.

src/reducer.ts
```diff
@@ -18,7 +18,7 @@ export type ReducerEntry<State, Payload = any> = [
   Reducer<State, Payload>
 ];
 
-type ReducerFn = {
+type ReducerCreator = {
```
holographic-principle: To add clarity after the parameter was renamed to avoid shadowing.

src/routines.ts
```diff
@@ -48,25 +48,25 @@ export const collectRoutines = (
...
+ * @param routineToSubscribe The routine to subscribe
+ * @param errorSubject An optional Subject that will receive errors form the
  *                     routine
  * @returns A Subscription object
  * @see defaultErrorSubject
  */
 export const subscribeRoutine = (
   action$: ActionStream,
-  routine: Routine<unknown>,
+  routineToSubscribe: Routine<unknown>,
```
holographic-principle: Shadowing.

totokaka:
Thanks! It's great to have common configs for this stuff.

PR #91, opened 2020-02-19
feat: Reducer constructor returns a reducer function

totokaka:
This commit makes the reducer constructor (`reducer`) return a function
that can be called like a plain reducer function. This allows for easier
composition of reducers.

BREAKING CHANGE: `ReducerEntry` has been renamed to `RegisteredReducer`,
and changed shape from an array to a function with a `trigger` property.

### DIFF
--- a/examples/reducers.spec.ts
+++ b/examples/reducers.spec.ts
@@ -1,4 +1,9 @@
-import { actionCreator, reducer, combineReducers, ReducerEntry } from 'rxbeach';
+import {
+  actionCreator,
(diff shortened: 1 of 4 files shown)


totokaka:
Heres a small usage example:
```ts
const handleIncrement = reducer(increment, (state: number) => state + 1);
const handleDoubleIncrement = reducer(doubleIncrement, (state: number) => handleIncrement(handleIncrement(state)));
```
The point is that you can directly use the first reducer inside the second

PR #95, opened 2020-02-20
Reduce state from arbitrary streams

totokaka:
This is what we have called "data reducers" in ardoq-front. I think everything should be clear from the commit messages and the code.

### DIFF
--- a/src/reducer.spec.ts
+++ b/src/reducer.spec.ts
@@ -19,15 +19,22 @@ const handleMany = reducer(
 const handleDecrement = reducer(decrement, throwErrorFn);
 const alwaysReset = reducer([incrementOne, incrementMany, decrement], () => 5);
 
-const inputs = {
+const actions = {
   '1': incrementOne(),
   '2': incrementMany(2),
   d: decrement(),
 };
-const outputs = {
+const words = {
+  a: '1',
+  b: '12',
+};
+const numbers = {
+  '1': 1,
   '2': 2,
+  '3': 3,
   '4': 4,
   '5': 5,
+  '6': 6,
 };
 
 test('reducer should store reducer function', t => {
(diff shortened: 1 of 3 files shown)


src/reducer.spec.ts
```diff
@@ -39,35 +44,86 @@ test('reducer should store reducer function', t => {
...
-    m.expect(source.pipe(combineReducers(1, [alwaysReset]))).toBeObservable(
-      expected
+    m.expect(action$.pipe(combineReducers(1, [alwaysReset]))).toBeObservable(
+      expected$
     );
   })
 );
+
+test(
+  'combineReducers should not emit before after the seed stream does',
```
anton164:
```suggestion
  'combineReducers should not emit and ignore actions until the seed stream emits a value',
```

^ would this make it a bit more clear?

src/reducer.ts
```diff
@@ -108,50 +139,120 @@ type ReducerCreator = {
...
+  type InitializedAccumulator = {
+    initialized: true;
+    reduced: boolean;
+    state: State;
+  };
+  type Accumulator = { initialized: false } | InitializedAccumulator;
+
+  const seed$ = wrapInObservable(seed).pipe(
+    take(1),
+    map(payload => [SEED_ORIGIN, payload] as ScanInput)
```
anton164:
```suggestion
    map(initialState => [SEED_ORIGIN, initialState] as ScanInput)
```

src/reducer.ts
```diff
@@ -108,50 +139,120 @@ type ReducerCreator = {
...
+        return {
+          initialized: true,
+          reduced: true,
+          state: reducerFn(accumulator.state, packet[1]),
+        };
+      },
+      { initialized: false } as Accumulator
+    ),
+    filter(
+      (acc): acc is InitializedAccumulator => acc.initialized && acc.reduced
```
anton164: Does this mean that the combine reducers stream doesn't emit until it's reduced once?

Shouldn't it also emit once it's been initialized?

src/reducer.ts
```diff
@@ -108,50 +139,120 @@ type ReducerCreator = {
...
-      const RegisteredReducer = reducersByActionType.get(type);
-      if (RegisteredReducer === undefined) {
-        // This shouldn't be possible
-        return state;
-      }
+    ofType(...actionReducers.flatMap(reducerFn => reducerFn.trigger.actions)),
+    map(action => [ACTION_ORIGIN, action] as ScanInput),
+    merge(seed$, ...source$s),
+    scan(
+      (accumulator, packet) => {
```
anton164: Very elegant - reads easily! 👍

src/reducer.ts
```diff
@@ -108,50 +139,120 @@ type ReducerCreator = {
...
     reducer: Reducer<State, VoidPayload>
   ): RegisteredReducer<State, VoidPayload>;
+
+  <State, Payload>(
+    source$: Observable<Payload>,
+    reducer: Reducer<State, Payload>
+  ): RegisteredReducer<State, Payload>;
 };
 
 export const reducer: ReducerCreator = <State>(
```
anton164: Great idea to overload this and just treat this as two types of reducers with either an action or a stream payload :)

src/reducer.ts
```diff
@@ -1,27 +1,58 @@
...
 
+const isActionReducer = <State, Payload>(
+  reducerFn: RegisteredReducer<State, Payload>
+): reducerFn is RegisteredActionReducer<State, Payload> =>
+  'actions' in reducerFn.trigger;
+
+const isStreamReducer = <State, Payload>(
+  reducerFn: RegisteredReducer<State, Payload>
+): reducerFn is RegisteredStreamReducer<State, Payload> =>
+  'source$' in reducerFn.trigger;
```
chriskr: Use const for 'actions' and 'sources$' ?

src/reducer.ts
```diff
@@ -108,50 +139,120 @@ type ReducerCreator = {
...
-        // This shouldn't be possible
-        return state;
-      }
+    ofType(...actionReducers.flatMap(reducerFn => reducerFn.trigger.actions)),
+    map(action => [ACTION_ORIGIN, action] as ScanInput),
+    merge(seed$, ...source$s),
+    scan(
+      (accumulator, packet) => {
+        if (packet[0] === SEED_ORIGIN) {
+          return { initialized: true, reduced: false, state: packet[1] };
```
chriskr: If if this is internal I think it would make sense to destruct packet to named variables

src/reducer.ts
chriskr: Yes, I would think the same?

Ok, I see, that correspond to how scan works.

src/reducer.ts
```diff
@@ -108,50 +139,120 @@ type ReducerCreator = {
...
  *
- * This operator does not change whether the stream is hot or cold.
+ * The behaviour is undefined if multiple reducers are registered for the same
+ * actions.
+ *
+ * This operator will only emit on relevant actions or if one of the source
+ * streams emits. Specifically, if the seed is a stream, this operator will not
+ * emit when the seed stream emits, but the first time an action of source
+ * stream emits after the seed stream has emitted. The seed stream will be
+ * unsubscribed after its first emit.
```
chriskr: At the moment I don't se the use case of seed as stream, do we already use this in our app?

This 

> Specifically, if the seed is a stream, this operator will not
   emit when the seed stream emits, but the first time an action of source
   stream emits after the seed stream has emitted.

Seems not entirely trivial, what is the motivation?

src/reducer.ts
```diff
@@ -108,50 +139,126 @@ type ReducerCreator = {
...
-        // This shouldn't be possible
-        return state;
-      }
+    ofType(...actionReducers.flatMap(reducerFn => reducerFn.trigger.actions)),
+    map((action): Packet => ({ source: ACTION_ORIGIN, value: action })),
+    merge(seed$, ...source$s),
+    scan(
+      (accumulator, packet) => {
+        if (packet.source === SEED_ORIGIN) {
+          return { initialized: true, reduced: false, state: packet.value };
```
chriskr: `reduced` is no longer needed?

totokaka: Good spot

chriskr:
> 1. Agreed. I've removed this check, so now the seed will be emitted once it's ready. This is a breaking change when seed is not a stream, as previously that would not be emitted

I took a closer look at the table view. Currently nextState is always called twice when opening that view, one from the kickstart and one from the combined sources stream. I think that is not ideal, we saw that nextState can be more expensive than expected. For most of the views we took care that valid state gets calculated at the moment of connecting the view, it seems now we will do this twice?

totokaka:
> > 1. Agreed. I've removed this check, so now the seed will be emitted once it's ready. This is a breaking change when seed is not a stream, as previously that would not be emitted
> 
> I took a closer look at the table view. Currently nextState is always called twice when opening that view, one from the kickstart and one from the combined sources stream. I think that is not ideal, we saw that nextState can be more expensive than expected. For most of the views we took care that valid state gets calculated at the moment of connecting the view, it seems now we will do this twice?

Whether we will emit multiple states right away depends on the reducers. In the table view I think we have the issue that our data reducers will emit on subscription, because the streams they depend on have `shareReplay`. This means that once the reducer is "attached" the streams will emit a value and the reducer will be invoked. 

We can solve this in three ways, I think:
  1. Not emit the seed from `combineReducers`
  2. Skip the replayed elements from the source streams of the reducers
  3. Debounce after `combineReducers`

Now, from the review, and in my opinion, `combineReducers` should emit the seed, as that seems most intuitive. Debouncing the stream means the state will still be calculated, but not  emitted, which is a workaround. So I guess the most apropriate solution is to skip the replayed elements on the reducers.

anton164:
Removing myself as reviewer for now. Feel free to re-request it's ready with the changes we discussed

src/reducer.spec.ts
```diff
@@ -39,35 +46,94 @@ test('reducer should store reducer function', t => {
...
+      expected$
+    );
+  })
+);
+
+test(
+  'combineReducers should reduce in predictable order',
+  marbles(m => {
+    // The "predictable order" seems to be the order in which the streams emit.
+    // Eg. The ordering of these seed definitions matter, but not the order of
```
anton164: I'm quite surprised to see that the order of the stream definitions matter 🤔

PR #99, opened 2020-02-28
ARD-4020 Implemented state stream tooling with tests (WIP)

anton164:
The pattern for creating state streams is one of the few missing pieces in ardoq-front to fully make use of reducers from rxbeach, so I thought I would give this a go.

The goal is for this to replace `createReducedStateStream` in ardoq-front (and hopefully most cases of `createStateStream` now that we have data reducers).

### The proposal

  **reduceState guarantees that:**
  - a value is emitted upon subscription with startWith (state streams can't be lazy)
  - the stream producer is shared, if there are multiple subscriptions
    the reducers will only run once
  - the state stream producer lives on "forever" throughout the app lifecycle,
    subsequent subscriptions will always get the previous state calculation
    even though there are no subscribers (i.e. the stream is not refcounted)

  **WIP:**
  - a state stream forwards errors to an error subject and continues emitting values after an error occurs

_All of the above "properties" have unit tests in **reduceState.spec.ts**_

### Example

With this tooling you can define a state stream like so:
```typescript
const defaultState = 1;
const handleOne = reducer(incrementOne, incrementOneHandler);
const handleMany = reducer(
  incrementMany,
  (accumulator: number, increment) => accumulator + increment
);
const state$ = action$.pipe(
  reduceState('testStream', defaultState, [
    handleOne,
    handleMany,
  ])
);
```

### DIFF
--- a/src/reduceState.spec.ts
+++ b/src/reduceState.spec.ts
@@ -1,0 +1,149 @@
+import { reducer, actionCreator } from 'rxbeach';
+import untypedTest from 'ava';
+import { marbles } from 'rxjs-marbles/ava';
+import sinon from 'sinon';
+import { reduceState } from 'rxbeach/reduceState';
+import { of, Subject } from 'rxjs';
+import { stubRethrowErrorGlobally } from 'rxbeach/internal/testUtils';
+
+const throwErrorFn = (): number => {
+  throw errors.e;
+};
+const incrementOne = actionCreator('[increment] one');
+const decrement = actionCreator('[increment] decrement');
+const incrementMany = actionCreator<number>('[increment] many');
+
+const incrementOneHandler = sinon.spy((accumulator: number) => accumulator + 1);
+const handleOne = reducer(incrementOne, incrementOneHandler);
+const test = stubRethrowErrorGlobally(untypedTest);
+const handleDecrementWithError = reducer(decrement, throwErrorFn);
+
+const actions = {
+  '1': incrementOne(),
+  '2': incrementMany(2),
+  d: decrement(),
+};
+const words = {
+  a: '1',
+  b: '12',
+};
+const numbers = {
+  '1': 1,
+  '2': 2,
+  '3': 3,
+  '4': 4,
+  '5': 5,
+  '6': 6,
+  '7': 7,
+};
+const errors = {
+  e: 'error',
+};
+
+test(
+  'reduceState should subscribe to action$ and emit initial state synchronously upon subscription',
+  marbles(m => {
+    const defaultState = 3;
+    const action$ = m.hot('  -');
+    const expected$ = m.hot('3', numbers);
+    const state$ = action$.pipe(reduceState('testStream', defaultState, []));
+
+    m.expect(action$).toHaveSubscriptions(['^']);
+    m.expect(state$).toBeObservable(expected$);
+  })
+);
+
+test(
+  'reduceState should only emit the latest state processed at the current frame',
+  marbles(m => {
+    const action$ = m.hot('  --1-', actions);
+    const word$ = m.hot('    a-a ', words);
+    const expected$ = m.hot('2-4', numbers);
+    const defaultState = 1;
+
+    const handleWord = reducer(
+      word$,
+      (state: number, word) => state + word.length
+    );
+    const state$ = action$.pipe(
+      reduceState('testStream', defaultState, [handleOne, handleWord])
+    );
+
+    m.expect(state$).toBeObservable(expected$);
+  })
+);
+
+test('reduceState should call reducer once when there are multiple subs', async t => {
+  const defaultState = 1;
+  incrementOneHandler.resetHistory();
+  const action$ = of(incrementOne());
+  const state$ = action$.pipe(
+    reduceState('testStream', defaultState, [handleOne])
+  );
+  const sub1 = state$.subscribe();
+  const sub2 = state$.subscribe();
+  t.assert(incrementOneHandler.calledOnce);
+  sub1.unsubscribe();
+  sub2.unsubscribe();
+});
+
+test(
+  'reduceState should share state and not reset as long as it has one subscriber',
+  marbles(m => {
+    const defaultState = 1;
+    const action$ = m.hot('      ----1-----1', actions);
+    const sub1 = '               --^--!';
+    const sub2 = '               ---------^--!';
+    const foreverSub = '         ^---------------';
+    const expected$ = m.hot('    1---2-----3-----', numbers);
+    const sub1Expected$ = m.hot('--1-2-----------', numbers);
+    const sub2Expected$ = m.hot('---------23-----', numbers);
+    const state$ = action$.pipe(
+      reduceState('testStream', defaultState, [handleOne])
+    );
+
+    m.expect(state$, foreverSub).toBeObservable(expected$);
+    m.expect(state$, sub1).toBeObservable(sub1Expected$);
+    m.expect(state$, sub2).toBeObservable(sub2Expected$);
+  })
+);
+
+test(
+  'reduceState should reset state when it has no subscribers',
+  marbles(m => {
+    const defaultState = 1;
+    const action$ = m.hot('      1--1---1---1-', actions);
+    const sub1 = '               --^--!--------';
+    const sub2 = '               ---------^---!';
+    const sub1Expected$ = m.hot('--12----------', numbers);
+    const sub2Expected$ = m.hot('---------1-2-', numbers);
+    const state$ = action$.pipe(
+      reduceState('testStream', defaultState, [handleOne])
+    );
+
+    m.expect(state$, sub1).toBeObservable(sub1Expected$);
+    m.expect(state$, sub2).toBeObservable(sub2Expected$);
+  })
+);
+
+test(
+  'reduceState catches errors and emits them to error subject',
+  marbles(m => {
+    const action$ = m.hot('  1d1', actions);
+    const expected$ = m.hot('2-3', numbers);
+    const errorMarbles = '   -e-';
+    const error$ = new Subject<any>();
+
+    m.expect(error$).toBeObservable(errorMarbles, errors);
+    m.expect(
+      action$.pipe(
+        reduceState(
+          'testStream',
+          1,
+          [handleOne, handleDecrementWithError],
+          error$
+        )
+      )
+    ).toBeObservable(expected$);
+  })
+);
--- a/src/reduceState.ts
+++ b/src/reduceState.ts
@@ -1,0 +1,45 @@
+import { markName, UnknownAction, defaultErrorSubject } from 'rxbeach/internal';
+import { tag } from 'rxjs-spy/operators';
+import { shareReplay, startWith, debounceTime } from 'rxjs/operators';
+import { combineReducers, RegisteredReducer } from 'rxbeach/reducer';
+import { OperatorFunction, pipe, Subject } from 'rxjs';
+
+/**
+ * Create a state stream for a set of reducers.
+ *
+ * It is guaranteed that the state stream will always emit a value upon subscription.
+ *
+ * The state stream is ref counted, which means that the stream
+ * will reset to its `defaultState` when there are no subscribers.
+ *
+ * If you wish to persist the state throughout the application lifecycle,
+ * you should create a subscriber that never unsubscribes.
+ *
+ * The values emitted from the stream are shared between the subscribers,
+ * and the reducers are only ran once per input action.
+ *
+ * @param name A name for debugging purposes
+ * @param action$ The action stream
+ * @param defaultState The initial state of the state stream,
+ *                     which is emitted synchronously upon subscription
+ * @param reducers The reducer entries that are combined with `combineReducers`
+ * @see rxbeach.combineReducers
+ * @returns An stream that emits the reduced state
+ */
+export const reduceState = <State>(
+  name: string,
+  defaultState: State,
+  reducers: RegisteredReducer<State, any>[],
+  errorSubject: Subject<any> = defaultErrorSubject
+): OperatorFunction<UnknownAction, State> =>
+  pipe(
+    combineReducers(defaultState, reducers, errorSubject),
+    startWith(defaultState),
+    shareReplay({
+      refCount: true,
+      bufferSize: 1,
+    }),
+    debounceTime(0),
+    markName(name),
+    tag(name)
+  );
--- a/src/reducer.spec.ts
+++ b/src/reducer.spec.ts
@@ -17,7 +17,7 @@ const handleMany = reducer(
   incrementMany,
   (accumulator: number, increment) => accumulator + increment
 );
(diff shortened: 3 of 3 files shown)


src/stateStream.ts
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @param name A name for debugging purposes
+ * @param action$ The action stream
+ * @param defaultState The initial state of the state stream,
+ *                     which is emitted synchronously upon subscription
+ * @param reducers The reducer entries that are combined with `combineReducers`
+ * @see rxbeach.combineReducers
+ * @returns An stream that emits the reduced state
+ */
+export const stateStream = <State>(
```
anton164: Not 100% settled on this signature. Since there are 4 args, perhaps it's better to take an object? In the end I did not use an object to stay consistent with the rest of rxbeach and since none of the arguments are optinoal

src/stateStream.ts
```diff
@@ -0,0 +1,82 @@
...
+    startWith(defaultState),
+    shareReplay({
+      refCount: false,
+      bufferSize: 1,
+    }),
+    markName(name),
+    tag(name)
+  );
+
+export const stateStreamWithErrorHandling = <State>(
```
anton164: This is obviously WIP, as I would like for this error handling to be a part of `stateStream`, however it turned out to be quite tricky to implement the desired behaviour (see the last test).

src/stateStream.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    );
+    const state$ = stateStream('testStream', action$, defaultState, [
+      handleWord,
+    ]);
+
+    m.expect(state$).toBeObservable(expected$);
+  })
+);
+
+test('stateStream subscribers should only see one emission if reducers emit synchronously', async t => {
```
anton164: This is a "good" side-effect, but per @chriskr 's suggestion, we should understand exactly why it happens 😅

src/stateStream.ts
```diff
@@ -0,0 +1,82 @@
...
+import { combineReducers, RegisteredReducer } from 'rxbeach/reducer';
+import { Subject, of } from 'rxjs';
+
+/**
+ * Create a state stream for a set of reducers.
+ *
+ * The state stream will always emit its last value upon subscription.
+ * On the first subscription to the state stream, it will emit `defaultState`.
+ *
+ * The state stream is not ref counted, which means that the underlying
```
chriskr: That means the stream never stops, continues to calculate state no matter if the state is used or not? Is this meant to be used for view state streams? Seems the wrong pattern to me? A view state stream should always calculate the valid state at the moment of connect?

src/stateStream.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    state$.subscribe();
+    m.expect(action$).toHaveSubscriptions(['^']);
+  })
+);
+
+test(
+  'stateStream should emit initial state synchronously upon subscription',
+  marbles(m => {
+    const defaultState = 3;
+    const action$ = m.hot('');
```
totokaka:
```suggestion
    const action$ = m.hot('  -');
```
To explicitly indicate it doesn't emit anything in the time frame the other stream emits `3`

src/stateStream.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    const expected$ = m.hot('3', numbers);
+    const state$ = stateStream('testStream', action$, defaultState, []);
+
+    m.expect(state$).toBeObservable(expected$);
+  })
+);
+
+test(
+  'stateStream should call reducers synchronously if they emit at once',
```
totokaka: I don't think this test belongs here - as it's just testing the functionality of `combineReducers`

src/stateStream.spec.ts
totokaka: I think this test is misguided. `toPromise` resolves once the stream completes. In this test `state$` will complete when `action$` completes.
I would guess `state$` actually emits the `defaultState`, but it's not caught because of the behavior of `toPromise`.

Anyways, let's not use `toPromise` in the tests :)

src/stateStream.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+);
+
+test('stateStream subscribers should only see one emission if reducers emit synchronously', async t => {
+  const defaultState = 1;
+  const action$ = of(incrementOne());
+  const state$ = stateStream('testStream', action$, defaultState, [handleOne]);
+  t.is(await state$.toPromise(), 2);
+});
+
+test('stateStream should call reducer once when there are multiple subs', async t => {
```
totokaka:
```suggestion
test('stateStream should call reducer once when there are multiple subs', t => {
```
An alternative way to test this would be to do it like the test `stateStream subscribes action$`, but check that `action$` still only have one subscription when the `state$` has two.

src/stateStream.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  const action$ = of(incrementOne());
+  const state$ = stateStream('testStream', action$, defaultState, [handleOne]);
+  const sub1 = state$.subscribe();
+  const sub2 = state$.subscribe();
+  t.assert(incrementOneHandler.calledOnce);
+  sub1.unsubscribe();
+  sub2.unsubscribe();
+});
+
+test('stateStream should never reset to its default state', async t => {
```
totokaka: I'm a bit unsure about how this test is implemented. We should look into if and how it is possible to test this behavior with marble tests.

src/stateStream.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  marbles(m => {
+    const action$ = m.hot('  -22--1--1', actions);
+    const word$ = m.hot('    ---a-b', words);
+    const expected$ = m.hot('135--6--7', numbers);
+    const defaultState = 1;
+
+    const handleWord = reducer(word$, () => {
+      throw 'error';
+      // eslint-disable-next-line no-unreachable
+      return 1;
```
totokaka:
```suggestion
    const handleWord = reducer(word$, (): number => {
      throw 'error';
```
I think this would work?

And this test would also be a bit more concise if we didn't use stream reducers (`handleWord`), but instead followed the style of this test: https://github.com/ardoq/rxbeach/blob/master/src/reducer.spec.ts#L95

src/stateStream.ts
totokaka: I believe a better signature would be `<State>(name: string, defaultState: State, reducers: RegisteredReducer<State, any>[]) => OperatorFunction<UnknownAction, State>`. This of course means this function is a stream operator.

Having this function be a stream operator is more composable, and will make it much nicer to use it together with `useNamespace`, for instance. It also avoids the problem of too many arguments.

src/stateStream.ts
totokaka: Yeah, this implementation is horrendous :fearful: 

I'll give it a spin and see if I can do something better. 

I hope the plan is to merge this and the `stateStream` function?

src/stateStream.ts
totokaka: This is a very good point, and warrants a proper discussion. I agree that the "view state stream" should be destroyed once the view is unmounted. If something should persist when the view is unmounted, it should go in another stream.

src/stateStream.ts
anton164: > I hope the plan is to merge this and the stateStream function?
Yes :) I split it out because it became so ugly 😅 Error handling and continuing, while emitting the correct things is a bit tricky with observables.

src/stateStream.ts
anton164: Gave this a bit more thought and I agree :) 

My initial "fear" was that making it composable would allow someone to misuse it, but I think we should aim to avoid misuse in other ways. Furthermore, your example of `useNamepace` clearly demonstrates why making it composable as an operator function is important, and I can imagine there might be other uses too.

src/stateStream.spec.ts
anton164: good suggestion, thanks 👍

src/stateStream.spec.ts
anton164: True, but I think this way is more explicit. I could add the test that ensures `action$` is only subscribed once, but that might be superfluous?

src/stateStream.spec.ts
anton164: True, it's not just misguided, it's completely at odds with the earlier test `reduceState should call reducers synchronously if they emit at once` 🤦‍♂ 

Removed now

src/stateStream.spec.ts
anton164: I gave up the first time, but finally found out how to do it :)

The key was to pass a subscriber as a second argument to `m.expect`

Relevant: [RTFM 😄  ](https://ncjamieson.com/marble-testing-rtfm/)

src/stateStream.spec.ts
anton164: Hmm, not sure if I agree because it is testing the combination of `startWith` and `combineReducers`. 

I renamed the test to: `reduceState should emit defaultState first and call reducers synchronously if they emit at once`

anton164:
@davidnegley / @chriskr  / @totokaka  thanks for the reviews!

@totokaka's discussion points:
> The name - I believe we shouldn't reserve the term "state stream" for reduced streams. Is "reduced state stream" a better name? At least it means we can avoid conflicts if we would like to add concepts with names like "derived state stream" later.
> 
> The signature - I strongly believe having this as an operator function is the way to go. See my arguments in the comments.

I agree to both points. I've renamed it to `reduceState` and refactored it to be an operator function.

> Two smaller thing:
>
> It seems the concept of incrementing has caught on as a nice test reducer. Maybe we should move these reducers, their handler functions and the actions, words and numbers object to their own file? It should probably be a separate PR or at least a different commit.

Agree, I can do a folllow-up once this is merged

> If the commit message starts with feat(state stream): we can get nice change logs :) I'm also unsure if it gives any value to have our internal issue number in the commit message.

Agree 👍 

I've addressed the rest of your feedback in the threads and/or via fixup commits.

src/stateStream.ts
anton164: For sure a good point! I think we should have a follow-up discussion on Wednesday.

The reasoning behind this decision was that I would like to try to avoid introducing an [isTemporary](https://github.com/ardoq/ardoq-front/blob/ab3dc69071c59595669173df6ac975d565801497/src/js/ardoq/streams/utils/stateStreamUtils.ts#L31) flag akin to what we have in `createStateStream` in ardoq-front.

From my observations in the codebase (and based on past experience with Redux), I think it's much easier to reason about state streams that aren't bound to a view's lifecycle, but rather the application's lifecycle. 

Three somewhat orthogonal reasons for this:
- **persisting state beyond a view's lifecycle:** a view might at some point unmount unexpectedly, but that doesn't necessarily mean that we wish to reset the state. In many cases, the best user experience is when the view state persist beyond the view's lifecycle. For example when you are editing a gremlin query, but click to open the workspaces module, then go back to editing the gremlin query.
- **reducing state for the view when the view is not mounted:** in some cases, a state stream should be ready to reduce actions before the view is mounted. I.e. we might wish to switch to a view within a different module and set some state. If we need to wait until the view is mounted to set the state, it can lead to some nasty code.
- **consistency and simplicity:** if we can avoid having two types of reduced state streams, one that resets and another that persists throughout the application lifecycle, I think that will simplify application development.

The first problem can be solved using **refCount: false**, while the second problem needs some kind of [global registry that sets up "global"/app-wide state streams](https://github.com/ardoq/ardoq-front/blob/dfe5c7312bd041858fb85a59e800a19647f7eccf/src/js/ardoq/startStateStreams.ts), similar to what is necessary for routines.

As both of you point out, there is also a third use-case: creating reduced state that is bound to a view, which only sets up once the view is mounted and completes once the view unmounts. This is especially true for our visualizations, since we don't want them to spend time building their view models when they are inactive.

However, here we must be careful to not mix two things - **reduced state** and **view model calculation**. 

IMO, view model calculation should obviously be bound to when a view mounts, since it is only necessary for the view when it's active, while **reduced state** can and should be bound to the application's lifecycle.

It's _highly possible_ that there's some important points that I'm missing, which requires us to have reducers that are bound to a view. That's why I think we should discuss this further :)

totokaka: Here's a problem case:
> The view model streams of visualizations should be reduced streams, have the same life-cycle as their view (React) components and be namespaced to each mounted instance. 

In this case the only thing that is in the WIP implementation of `reduceState` that we don't need / don't want is `shareReplay`.

I think one take-away from this is that we don't really have a use case where we need `shareReplay` without `refCount: false`. So maybe we could define a very simple alias for `shareReplay({ refCount: false, bufferSize: 1 })` named something like `persistAndReplay`? Then reduced state streams would need two operators, but we're free to reuse `reduceState` when creating the tooling for view model streams.

src/reduceState.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    const expected$ = m.hot('    1---2-----3-----', numbers);
+    const sub1Expected$ = m.hot('--1-2-----------', numbers);
+    const sub2Expected$ = m.hot('---------23-----', numbers);
+    const state$ = action$.pipe(
+      reduceState('testStream', defaultState, [handleOne])
+    );
+
+    m.expect(state$).toBeObservable(expected$);
+    m.expect(state$, sub1).toBeObservable(sub1Expected$);
+    m.expect(state$, sub2).toBeObservable(sub2Expected$);
```
totokaka: This really nice!

src/stateStream.ts
anton164: True, there are some obvious examples from our current views where we wish to have the reset pattern, but still keep reducers around 🤔 I was trying hard to entertain the idea that we can live without them, but that's a bit detached from reality 😄 

However, I don't think it's a good idea to add a `persistAndReplay` operator, since it would most require state stream definitions to be:
```typescript
const state$ = action$.pipe(
   reduceState(),
   persistAndReplay()
);
// instead of simply
const state$ = action$.pipe(
   reduceState()
);
```

IMO, we should either add some option to `reduceState` that allows to configure this behaviour, or create a separate interface, i.e. `reduceViewState`. If we roll with `reduceViewState` it might make sense to also rename `reduceState`.

anton164: Added another test, which shows an interesting consequence of `({ refCount: false })`:
https://github.com/ardoq/rxbeach/blob/231ad441add1b662c430a893affbe516b8b10e71/src/reduceState.spec.ts#L111-L127

Perhaps, the best solution is actually to roll with `refCount: true` instead. This allows for more fine-grained control as we can distinguish between "permanent" and "temporary" streams by having a registry. 

If you need a "local" stream that resets, you simply avoid registering it in `startGlobalStateStreams` registry.

If you need a "local" stream per view instance, you have a simple method:
```typescript
const createInstanceStream (action$, instanceId) => action$.pipe(
   withNamespace(instanceId),
   reduceState(...)
);
```

anton164: Finally, I have pushed a commit that switches the implementation to `refCount: true`. I think that's the best way to go :) I've requested a new review

PR #104, opened 2020-03-04
Export reduce state

andreajl:


anton164:
fYI @andreajl I rebased and reworded your commits.

We use conventional commits in RxBeach https://www.conventionalcommits.org/

PR #105, opened 2020-03-04
ARD-3813 add 10 stream parameter overload

davidnegley:
alphabet practice

trying to address this @ts-ignore https://github.com/ardoq/ardoq-front/blob/master/src/js/ardoq/streams/assets/assets$.ts#L249 for a call to `derivedStream` which just recently increased the number of streams passed to 10.


totokaka:
LGTM. Could you do the same for other functions we have with variable number of genetic parameters? Then we can just say that we should support up to 10 generic parameters in each of these functions.

davidnegley:
okay. i only found one, though. i was searching for <A>

PR #108, opened 2020-03-05
chore: lint commit messages on commit

tlaundal:
I want to try this for a while. We might find it too aggressive, or we might end up loving it :)

I'm most concerned about losing long commit messages because of silly mistakes. Say I write a paragraph explaining my changes, and then `commitlint` rejects the commit message because the header had a capital letter in the commit subject.

## Explanation of the most important rules:
```
chore(internal): lint commit messages on commit
```
- The header may begin with a `fixup!`
- After `fixup!` there must be a type. One of: 'build', 'ci', 'chore', 'docs', 'feat', 'fix', 'perf', 'refactor', 'revert', 'style', 'test'
- After the type there may be a scope. It must be all lower case, but may contain spaces, dashes or underscores
- The subject itself cannot have a capital first letter

Over time I think it will come naturally what scopes to use. The major thing about the scope is that it affects the changelog. It is completely fine to omit the scope also.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,11 +26,14 @@
   },
   "devDependencies": {
     "@ardoq/shared-configs": "^0.0.13",
+    "@commitlint/cli": "^8.3.5",
+    "@commitlint/config-conventional": "^8.3.4",
     "@istanbuljs/nyc-config-typescript": "1.0.1",
     "@types/sinon": "7.5.2",
     "ava": "3.5.0",
     "conditional-type-checks": "1.0.5",
     "coveralls": "3.0.9",
+    "husky": "^4.2.3",
     "nyc": "15.0.0",
     "rimraf": "3.0.2",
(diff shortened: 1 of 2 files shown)


tlaundal:
> Does this hook reject a commit message if not passing the linting?

Yes, it will just tell you what the problem is. If you provided the message through `git commit -m 'message'`, it's no issue, as you can just modify the command, but if you write the message in an editor, it will tell you what the issue is, without providing a way of modifying the rejected message, so you have to write it from scratch.

tlaundal:
> Surely someone must have made a fix for that problem?

I couldn't find anything, but it was a quite hard thing to google for, though, so maybe someone else will have more luck googling?

tlaundal:
I'll merge this. Let's revert it if it causes issues.

PR #109, opened 2020-03-05
Changelog

tlaundal:
Adds a changelog for the last couple of releases
Adds standard-version for bumping the version and generating the changelog.


tlaundal:
> Do you think this tool would spread to our other repos as well?

Definitively. The only thing it does is bumping the version and creating the changelog

PR #113, opened 2020-03-12
Use @ava/typescript for testing

tlaundal:
This PR moves us away from `ts-node` and `tsconfig-paths` in favour of using [@ava/typescript](https://github.com/avajs/typescript) for running our `.spec.ts` tests. This should be a simpler and more robust setup.

Some big changes had to be made along the way
- Deleted the `examples` directory. It provides too little value to be worth maintaining. It's function should be fulfilled by our documentation (which is in need of updating)
- Made all imports relative. There is generally poor support for absolute imports in the tooling around typescript. Doing this also reveals that we should do some work on our folder structure

With this change, you need to run both `yarn watch` and `yarn test:watch` in order to have live test running.

It's probably best to review commit by commit.


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "esnext",
+    "target": "es2019",
```
holographic-principle: Check that the coverage line numbers are correct with this setting (otherwise lowering the target might work). I seem to remember stumbling upon problems people have had in this regard.

package.json
```diff
@@ -17,22 +17,21 @@
...
   "devDependencies": {
     "@ardoq/shared-configs": "^0.0.13",
+    "@ava/typescript": "1.1.1",
     "@commitlint/cli": "^8.3.5",
     "@commitlint/config-conventional": "^8.3.4",
-    "@istanbuljs/nyc-config-typescript": "1.0.1",
     "@types/sinon": "7.5.2",
     "ava": "3.5.0",
     "conditional-type-checks": "1.0.5",
-    "coveralls": "3.0.9",
```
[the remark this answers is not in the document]
tlaundal: Done on purpose. We don't need `coveralls` in our local repos. `npx coveralls` will install it as needed in CI.

tsconfig.json
tlaundal: Did a quick test, and they seem to work as expected.

tlaundal:
Sorry for the fixup party. Just trying to make the CI build as smooth as possible. Latest few fixups was to include coveralls in the cache.

.circleci/config.yml
```diff
@@ -15,21 +15,18 @@ jobs:
...
+            - v4-dependencies-{{ checksum "yarn.lock" }}
+            - v4-dependencies-
 
-      - run: yarn install
-
-      - save_cache:
-          paths:
-            - node_modules
-          key: v1-dependencies-{{ checksum "package.json" }}
+      - run: yarn install --frozen-lockfile
```
andreajl: It might do the trick to avoid this change to do a `yarn install --force` locally and commit the new lockfile. Also might be worth it to remove all the hats `^` from the package file to avoid "wrong" dependencies.

Sorry for being 🐌 to apply the orb.

tlaundal: The reason I added `--frozen-lockfile` is that it will fail the build if `yarn.lock` and `package.json` do not correspond. 

We should remove the hats from all our normal and dev dependencies, but they should stay in the peer dependencies.

tlaundal:
See avajs/typescript#3 and avajs/typescript#12 for progress on ava also running the compilation.

I'll go ahead and merge this since there have been no objections :)

tlaundal:
This build failure is a perfect example of why `--frozen-lockfile` might be useful!
https://app.circleci.com/pipelines/github/ardoq/rxbeach/276/workflows/fcebbd45-99ae-4bce-8a52-f868688c5b58/jobs/273

PR #115, opened 2020-03-19
feat(react): add useState React hook

tlaundal:
Documentation on the function and tests should tell you everything you need to know.

I was contemplating splitting this out to a different package, so as to avoid adding react as a dependency to RxBeach itself. But I guess that's something to consider much later.

### DIFF
--- a/src/react/useStream.ts
+++ b/src/react/useStream.ts
@@ -10,8 +10,21 @@ import { Observable } from 'rxjs';
  *
  * This hook passes the `source$` argument as a dependency to `useEffect`, which
  * means you will need to take care that it is referentially equal between each
- * render (unless you want to resubscribe, of course). Generally, you should
- * only use this hook for static/global streams.
+ * render (unless you want to resubscribe, of course).
+ *
+ * For using this hook with streams defined (created, piped, etc.) inside a
+ * component, you need to wrap the stream definition in `useMemo`:
+ * ```tsx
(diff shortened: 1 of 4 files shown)


src/react/useStream.ts
```diff
@@ -0,0 +1,43 @@
...
+ *   return <p> Latest payload: { payload } </p>;
+ * }
+ * ```
+ *
+ * @param source$ Stream that provides the needed values
+ * @param initial Initial value to return
+ * @see useEffect
+ */
+export const useStream = <T, I>(source$: Observable<T>, initial: I): T | I => {
+  const [value, setValue] = useState<T | I>(initial);
```
anton164: An observation from using a similar implementation of the hooks pattern a bit in one of our repos. When using the `useStream` hook think it's a bit annoying that you have to deal with the `initial` value within the view.

Normally when using `connect` to connect to a stream, the view doesn't render until the stream emits a value, which means that you don't have to deal with any intermediate value unless you choose to.

With this hooks pattern, you need to handle the `initial` value, although in practice what you actually want is to wait for the stream to emit or get the initial value from the stream.

Not sure if this can be done differently, but I think it has some implications for how we write views.

tlaundal:
I've now removed the `initial` argument, and instead made the hook return the symbol `NOT_YET_EMITTED`. For the HOC we can do as before, and just not draw the component before the a value is emitted.

PR #116, opened 2020-03-19
feat(react): add connect React HOC

tlaundal:
A higher order component for connecting react components to streams.
Builds upon the `useStream` hook, and is intended for use in class based components that can't use the hook themselves.

Rigid testing with `react-test-renderer`, which renders the components without a DOM. Have also chosen to not write any `tsx`, so we won't have to deal with more tooling.

### DIFF
--- a/src/react/connect.ts
+++ b/src/react/connect.ts
@@ -1,0 +1,34 @@
+import { useStream, NOT_YET_EMITTED } from './useStream';
+import { Observable } from 'rxjs';
+import React, { ComponentType } from 'react';
+
+/**
+ * Higher order component for connecting a component to a stream
+ *
+ * The stream does not need to provide all the properties the component expects,
+ * any missing properties will be required by the component returned from this
+ * component.
+ *
+ * Typescript does, in certain cases, allow for extra properties to be unpacked
+ * on a component. This means it may be possible to send props which would be
+ * provided by the stream. In these cases the values from the stream will
+ * override any values passed as props.
+ *
+ * @param WrappedComponent Component that will recieve props from the stream
+ * @param stream$ Stream that will provide props to the component
+ * @see useStream
+ */
+export const connect = <Props, Observed extends Partial<Props>>(
+  Component: ComponentType<Props>,
+  stream$: Observable<Observed>
+) => (props: Omit<Props, keyof Observed>) => {
+  const value = useStream(stream$);
+
+  if (value === NOT_YET_EMITTED) return null;
+
+  // Typescript doesn't recognize this as Observed & Props for some reason
+  // Question on StackOverflow: https://stackoverflow.com/q/60758084/1104307
+  const newProps = { ...props, ...value } as Props & Observed;
+
+  return React.createElement(Component, newProps);
+};
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "husky": "4.2.3",
     "nyc": "15.0.0",
     "react": "16.13.1",
-    "react-test-renderer": "16.13.0",
+    "react-test-renderer": "16.13.1",
     "rimraf": "3.0.2",
     "rxjs": "6.5.4",
     "rxjs-marbles": "6.0.0",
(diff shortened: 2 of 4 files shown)


src/react/connect.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export const connect = <Props, Observed extends Partial<Props>>(
+  Component: ComponentType<Props>,
+  stream$: Observable<Observed>
+) => (props: Omit<Props, keyof Observed>) => {
+  const value = useStream(stream$);
+
+  if (value === NOT_YET_EMITTED) return null;
+
+  // Typescript doesn't recognize this as Observed & Props for some reason
```
holographic-principle, line 24: If you use `Exclude` instead of `Omit`, does it work then?

src/react/connect.ts
```diff
@@ -0,0 +1,29 @@
...
+  // Question on StackOverflow: https://stackoverflow.com/q/60758084/1104307
+  const newProps = { ...props, ...value } as Props & Observed;
```
holographic-principle, line 26: Isn't the precedence in the current `connect` (from `streamUtils`) the opposite? (`...value, ...props`)

anton164: Nice catch @holographic-principle.

Personally, I don't see a strong case for why "props" or "streamedProps" should take precedence 🤷‍♂ Did you switch these around for any particular reason @tlaundal ? 

In general, I can't think of a case where having the same key defined in "props" and "streamedProps" is a good pattern as I think this could cause a lot of confusion. Perhaps we could add a warning in dev mode?

src/react/connect.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param WrappedComponent Component that will recieve props from the stream
+ * @param stream$ Stream that will provide props to the component
+ * @see useStream
+ */
+export const connect = <Props, Observed extends Partial<Props>>(
+  Component: ComponentType<Props>,
+  stream$: Observable<Observed>
+) => (props: Omit<Props, keyof Observed>) => {
+  const value = useStream(stream$);
```
anton164, line 20: I think I remember you mentioned that this would solve `connectWithProps` as well, do you have an example of that?

holographic-principle: I'm purely guessing here, but I think Tobias' intention might have been to replace `connectWithProps` with the HoC from this PR. That would explain the order in which props vs streamedProps are spread. However this is (potentially) not backwards-compatible with `connect`.

src/react/connect.ts
holographic-principle: Check if this can be applied: https://github.com/ardoq/shared-ui/blob/eb13e2f926a7db3a80e71ad5f657a9bfce976414/src/utils.tsx#L31-L49

tlaundal: `Omit` is a shorthand for the `Pick` and `Exclude` done on line 31 of your link.
The reason your example works, is that the added props are known at compile time.

See this SO answer: https://stackoverflow.com/a/56178198/1104307
And this issue in the TS bugtracker: https://github.com/microsoft/TypeScript/issues/36316

src/react/connect.ts
tlaundal: So actually, this kind of shadowing is actually caught by TS because of the types (this is what the `Omit` does), so I don't really think this is a problem. 

I can change the order to be the same as in ardoq-front, just to be on the safe side, but I don't think we should document or test this, as it is explicitly forbidden by the types.

src/react/connect.ts
tlaundal: `connectWithProps` use properties as a part of the stream (usually filtering). The `useStream` hook can be used in it's place.
This HOC cannot replace `connectWithProps` because the properties are not available when the stream is initialized.

I'll create a PR in ardoq-front where I replace `connectWithProps` with `useStream` today.

tlaundal:
I have clearified in the documentation and with tests the precendence of passed/forwarded props and streamed props. Will merge in about half an hour.

src/react/connect.ts
anton164: Ok, good 👍

PR #119, opened 2020-03-23
fix(reducers): ensure reduceState replays values to derived streams

anton164:
Apparently the order of debounceTime and shareReplay operators matter.
By moving debouncing before shareReplay, we ensure that reduceState
correctly replays values to derived streams (typically routines).

Added a comment that details reduceState debounced behaviour.

Also added a test that shows how reduceState sometimes might emit
outdated values to derived streams due to debouncing if the
derived stream is triggered in the same frame.

### DIFF
--- a/src/operators/reduceState.spec.ts
+++ b/src/operators/reduceState.spec.ts
@@ -5,6 +5,8 @@ import { reduceState } from '../operators/reduceState';
 import { of, Subject } from 'rxjs';
 import { stubRethrowErrorGlobally } from '../internal/testing/utils';
 import { incrementMocks } from '../internal/testing/mock';
+import { withLatestFrom, map } from 'rxjs/operators';
+import { ofType } from './operators';
 
 const { reducers, actionCreators, handlers } = incrementMocks;
 const { actions, words, numbers, errors } = incrementMocks.marbles;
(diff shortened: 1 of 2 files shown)


src/operators/reduceState.spec.ts
```diff
@@ -112,3 +114,69 @@ test(
...
+
+test(
+  'reduceState emits outdated value to derived streams due to debouncing',
+  marbles(m => {
+    const defaultState = 1;
+    const action$ = m.hot('      -2-(12)-2-', actions);
+    const sub1 = '               ^---------';
+    const sub1Expected$ = m.hot('1--2------', numbers);
+    const sub2 = '               -^--------';
+    const sub2Expected$ = m.hot('-1-1----2-', numbers);
```
chriskr: I think this level of complexity is gets to high. It's basically a result of mixing 'sync with async (or debounced) streams. Keeping track of this seems almost impossible in our app.

tlaundal:
Seems like a good fix, though I found it hard to understand the test "reduceState emits outdated value to derived streams due to debouncing".
I also agree with @chriskr that we should do some serious thinking on the complexity and behavior of this.

anton164:
> Seems like a good fix, though I found it hard to understand the test "reduceState emits outdated value to derived streams due to debouncing".
> I also agree with @chriskr that we should do some serious thinking on the complexity and behavior of this.

Perhaps could have been phrased better- suggestions are welcome! Hopefully we can get rid of that test once we get to the root of the problem though :)

We discussed with @chriskr yesterday that we perhaps need a `withLatestState` operator for combining state from state streams that are debounced with `reduceState`.

PR #121, opened 2020-03-26
fix(reducers): removed debouncing from reduceState

anton164:
Debouncing state streams made it unsafe to use withLatestFrom,
  since the stream calculations were debounced.

  For now, reverted to emitting all state calculations.

### DIFF
--- a/src/operators/reduceState.spec.ts
+++ b/src/operators/reduceState.spec.ts
@@ -28,11 +28,11 @@ test(
 );
 
 test(
-  'reduceState should only emit the latest state processed at the current frame',
+  'reduceState always emits defaultState and every subsequent calculation',
   marbles(m => {
-    const action$ = m.hot('  --1-', actions);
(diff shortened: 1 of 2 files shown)


src/operators/reduceState.spec.ts
```diff
@@ -180,3 +180,47 @@ test(
...
+      map(([, s]) => s)
+    );
+
+    m.expect(state$, sub1).toBeObservable(sub1Expected$);
+    m.expect(routine$, sub2).toBeObservable(sub2Expected$);
+  })
+);
+
+test.failing(
+  'reduceState should only emit the latest state processed at the current frame to prevent glitches',
```
chriskr: I think we should discuss the expectation here. This is not really a glitch, it's just two iterations in one frame.

anton164: True, I guess my intent behind the wording is that two iterations in one frame can lead to glitches further down the graph

PR #140, opened 2020-04-15
feat: allow listening for subscribed markers

tlaundal:
For our streams-to-ardoq functionality we need to have a sentral registry of streams / stream markers that can be used to build the graph of the streams. Previously we did this with the `streamRegistry` module in ardoq-front, which the stream tooling in `ardoq-front` hooked into.

With the introduction of `connect` and `useStream` in RxBeach, `ardoq-front` no longer has a means of registering streams, at least not without wrapping / decorating / replacing functions from RxBeach. 

Originally I wanted to keep RxBeach as free from global state as possible, which is why I came up with the hooks approach. It a simple callback-based event approach. I'm not at all sure about the naming. 

This PR adds a function for registering a callback to registered streams. The idea is that we could use this callback to build up the registry in `ardoq-front`. I am having some second thoughts now - maybe we should rather actually build the registry directly here in RxBeach?

### Added
- Markers for views and routines
- A simple hook system, for listening for subscribed streams
- hooks are invoked when routines are subscribed, or routines are connected to react views

### To investigate
- Can we "register" streams before they are subscribed?
- Will we manage to add a hook early enough in ardoq-front, to receive all the markers?
- We should add tooling for global subscriptions, so they are automatically subscribed.

### DIFF
--- a/src/hooks.spec.ts
+++ b/src/hooks.spec.ts
@@ -1,0 +1,38 @@
+import test from 'ava';
+import { clearHooks, hookMarkers, notifyStreamSubscribed } from './hooks';
+import { never } from 'rxjs';
+import { markName } from './internal';
+import { MarkerType } from './internal/markers';
+
+test.afterEach(() => clearHooks());
+
+test('marker hooks are invoked', (t) => {
+  const stream = never().pipe(markName('test'));
+
+  hookMarkers((marker) => {
+    t.deepEqual(marker.type, MarkerType.NAME);
+  });
+
+  notifyStreamSubscribed(stream);
+});
+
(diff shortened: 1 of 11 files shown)


holographic-principle:
@tlaundal can you give me some background on what we are trying to achieve?

tlaundal:
> @tlaundal can you give me some background on what we are trying to achieve (beyond the PR description)?

Yes, sorry for the lack on info. I've written a  longer description now, it should explain the motivation. Writing it got me thinking about whether this is actually a good way to solve the problem or not.

tlaundal:
I've changed approach to this. We will instead just make RxBeach expose a function `getStreamGraph`

PR #141, opened 2020-04-16
docs: added initial scaffolding for interactive RxBeach docs

anton164:
First draft can be seen live here:
https://ardoq.github.io/rxbeach/#/

The content needs some work, but the goal of this PR is to add the initial scaffolding and allow us to update the pages in subsequent PRs.


react-docs/src/SideMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+      defaultOpenKeys={pages
+        .filter(page => isPageGroup(page))
+        .map(page => page.title)}
+      selectedKeys={[selectedPage.id]}
+      mode="inline"
+    >
+      <h2>🏖️ RxBeach</h2>
+      {pages.map(page =>
+        isPageGroup(page) ? (
+          <SubMenu key={`${page.title}`} title={page.title}>
```
andreajl, line 41: `key={page.id}`

react-docs/src/pages/Actions.tsx
```diff
@@ -0,0 +1,252 @@
+import React from 'react';
+import CodePreview from './CodePreview';
+import { MainHeading } from './atoms';
+
+const Introduction = () => {
+  return (
+    <div>
+      <MainHeading>Actions</MainHeading>
```
andreajl, line 8: I probably miss some context, but why not use markdown for the text? I guess you have considered another option: By having all the documentation as `.md` files would make it easier to write and I would argue it scales better. The snippets could lie in the same folder as the text.

react-docs/webpack.config.js
```diff
@@ -0,0 +1,43 @@
...
+        use: ['babel-loader'],
+      },
+      {
+        test: /\.css$/i,
+        use: ['style-loader', 'css-loader'],
+      },
+    ],
+  },
+  resolve: {
+    extensions: ['.mjs', '.tsx', '.ts', '.js'],
```
andreajl, line 24: What are `.mjs`?

react-docs/src/pages/atoms.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { convertStringToId } from '../utils';
+
+export const MainHeading: React.FC = ({ children }) => {
```
andreajl, line 4: ```export const MainHeading = styled.h2` ```?

react-docs/src/pages/Actions.tsx
```diff
@@ -0,0 +1,252 @@
...
+          </p>
+          <p>
+            Furthermore, "describing events that occurred" can prevent coupling
+            between modules. A module should know how to describe an event that
+            happens through an action, but it shouldn't care about how that
+            action is executed (i.e. a certain order of action setters).
+          </p>
+        </li>
+        <li>
+          <h4>Payloads should only contain "pure" data</h4>
```
chriskr, line 137: I think you could summarize that as payloads should be 'jsonable' ?

react-docs/src/pages/Actions.tsx
```diff
@@ -0,0 +1,252 @@
...
+          <h4>
+            Payloads should be concise, avoid passing unnecessary information
+          </h4>
+          <p>
+            Actions should describe what's happening by passing as little
+            information as possible. This is in order to ensure that handlers of
+            the actions are reading data from the correct source of truth.
+          </p>
+          <p>
+            In practice, this means that you should always to aim pass ids
```
chriskr, line 161: 'you should always aim to pass' ?

react-docs/src/pages/Views.tsx
```diff
@@ -0,0 +1,202 @@
...
+import { take, map } from 'rxjs/operators';
+import { useStream, NOT_YET_EMITTED, connect } from 'rxbeach/react';
+
+const Views = () => {
+  return (
+    <div>
+      <MainHeading>Views</MainHeading>
+      <p>Goal: Simple views that react to state changes</p>
+      <p>
+        A view's core responsibility is to render HTML and dispatch events
```
chriskr, line 14: Strictly speaking we are not using HTML, just 'display data' ?  Or display data with web technologies? Technically it's more about creating a DOM?

react-docs/src/pages/Views.tsx
```diff
@@ -0,0 +1,202 @@
...
+        const View = ({ number }: ViewModel) => {
+          return <p>Value: {number}</p>;
+        };
+
+        const ConnectedView = connect(View, viewModel$);
+
+        render(<ConnectedView />);`}
+      />
+      <h3>
+        <code>useStream</code> - Hooks Example
```
chriskr, line 85: Should be removed now?

react-docs/src/pages/Actions.tsx
anton164: Yeah, I see your point! For now it is written in JSX because it possible to use the CodePreview component, and I believe all of us should be capable of reading and updating JSX easily. 

However as we discussed with @tlaundal, in a v2 of this we could probably move to markdown and find a way to "enhance" them with code previews. I think this is how [antd](https://raw.githubusercontent.com/ant-design/ant-design/master/components/alert/demo/error-boundary.md) produce their documentation. I first tried to see if we could re-use their approach/code, but it wasn't trivial.

PR #142, opened 2020-04-16
Reduced State Stream Tooling

tlaundal:
This PR investigates how we can do the final level of tooling for creating reduced state streams.

A reduced state stream is a stream built up of reducers (action- and stream reducers) that contains global application state. They should have only one instance each and never reset their state.

As the state is never reset, i.e. it is independent of subscriptions, it can also be exposed through a property on the stream. This gives us a nice interface between the stream and sync worlds, and could make `withLatestFrom` unnecessary in many cases (Though that might not be something we want to do). 

The current implementation in this PR extends `BehaviorSubject` and subscribes the stream when created. This is not ideal, as `next`, `error` and `complete` are available as methods on the objects, though they should never be used. It also means the state can only be exposed through the property name `value`, which is very non-descriptive.

The intended way to use this in ardoq-front is like this:
```ts
// in myState$.ts
const initialState = { ... };
const reducers = [ ... ];
const createMyState$ = (action$: ActionStream, other$: OtherStream) => {
  const handleOther = reducer(other$, ...);

  return createReducedStream(
    'myState$',
    initialState,
    [handleOther, ...],
    action$
  );
}

// in streams.ts
export const myState$ = createMyState$(action$, other$);
```
This approach makes it easy to test the stream, and makes the dependencies explicit.

I will create some comment threads for things I think we want to discuss.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -35,3 +35,5 @@ export {
 export { derivedStream } from './derivedStream';
 
 export { combineLatest, merge, zip } from './decoratedObservableCombiners';
+
+export { persistentReducedStream } from './persistentReducedStream';
--- a/src/persistentReducedStream.spec.ts
+++ b/src/persistentReducedStream.spec.ts
@@ -1,0 +1,193 @@
+import test from 'ava';
+import {
+  StateStream,
+  persistentReducedStream,
+} from './persistentReducedStream';
+import { never, of, Subject } from 'rxjs';
+import { marbles } from 'rxjs-marbles/ava';
+import { incrementMocks } from './internal/testing/mock';
+import { map } from 'rxjs/operators';
+
+const { reducers, actionCreators, handlers } = incrementMocks;
+const { actions, numbers, errors } = incrementMocks.marbles;
+const reducerArray = Object.values(reducers);
+
+test('StateStream should expose its state immediately', (t) => {
+  const state = 'hello';
+  const state$ = new StateStream(state, never());
+
+  t.deepEqual(state$.state, state);
+});
+
+test('StateStream should not initially be closed', (t) => {
+  const state$ = new StateStream(null, never());
+
+  t.false(state$.closed);
+});
+
+test(
+  'StateStream should follow the underlying observable',
+  marbles((m, t) => {
+    const source$ = m.hot('-abc');
+    const expected = '     0abc';
+
+    const state$ = new StateStream('0', source$);
+
+    m.expect(state$).toBeObservable(expected);
+
+    m.flush();
+    t.deepEqual(state$.state, 'c');
+  })
+);
+
+test(
+  'StateStream should error when the underlying observable errors',
+  marbles((m) => {
+    const source$ = m.hot('-a#');
+    const expected = '     0a#';
+
+    const state$ = new StateStream('0', source$);
+
+    m.expect(state$).toBeObservable(expected);
+  })
+);
+
+test(
+  'StateStream should complete when the underlying observable completes',
+  marbles((m) => {
+    const source$ = m.hot('-a|');
+    const expected = '     0a|';
+
+    const state$ = new StateStream('0', source$);
+
+    m.expect(state$).toBeObservable(expected);
+  })
+);
+
+test(
+  'StateStream should support piping',
+  marbles((m) => {
+    const source$ = m.hot('-12');
+    const expected = '     024';
+
+    const state$ = new StateStream(0, source$);
+    const actual$ = state$.pipe(map((a) => `${a * 2}`));
+
+    m.expect(actual$).toBeObservable(expected);
+  })
+);
+
+test(
+  'StateStream should be possible to unsubscribe',
+  marbles((m, t) => {
+    const trigger = m.hot('--|');
+    const source$ = m.hot('-a-c');
+    const expected$ = '    0a--';
+    const subscription = ' ^-!';
+
+    const state$ = new StateStream('0', source$);
+    trigger.subscribe({ complete: () => state$.unsubscribe() });
+
+    m.expect(state$).toBeObservable(expected$);
+    m.expect(source$).toHaveSubscriptions(subscription);
+
+    m.flush();
+    t.true(state$.closed);
+  })
+);
+
+test('persistentReducedStream should expose the state immediately', (t) => {
+  const initialState = 5;
+  const state$ = persistentReducedStream('test', initialState, [], never());
+
+  t.deepEqual(state$.state, initialState);
+});
+
+test(
+  'persistentReducedStream reduces state',
+  marbles((m, t) => {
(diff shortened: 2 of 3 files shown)


src/createReducedStream.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @param name The name of the stream, used for placing a marker and spy tag
+ * @param initialState The initial state of the stream
+ * @param reducers The reducers that build up the stream state
+ * @param action$ The action stream the action reducers should reduce over
+ */
+export const createReducedStream = <State>(
+  name: string,
+  initialState: State,
+  reducers: RegisteredReducer<State, any>[],
+  action$: ActionStream,
```
tlaundal, line 45: **Is it intuitive to have `action$` as the last required argument?**

The argument list here is somewhat long, but I think it's better to have them as normal arguments than to put them in an object. Besides I believe that once we merge this, we should not attempt to change the signature without making a breaking change, so we can reconsider from scratch if we need to add more arguments later.

src/createReducedStream.ts
```diff
@@ -0,0 +1,54 @@
...
+  errorSubject: Subject<any> = defaultErrorSubject
+): ReducedStream<State> => {
+  const reducedState$ = action$.pipe(
+    combineReducers(initialState, reducers, errorSubject),
+    markName(name),
+    tag(name)
```
tlaundal, line 51: **This is copied from `reduceState`**, but does not contain `shareReplay` or `startWith`, because the `ReducedStream` handles that.

Do we still have a need for both? As of now I would say yes, because we need to use `reduceState` directly where we use namespaces, but once we have high level tooling for namespaces, I think the need will dissapear.

src/createReducedStream.ts
```diff
@@ -0,0 +1,54 @@
...
+import { markName } from './internal/markers';
+import { tag } from 'rxjs-spy/operators';
+
+export class ReducedStream<State> extends BehaviorSubject<State> {
+  subscription: Subscription;
+
+  constructor(initialState: State, reducedState$: Observable<State>) {
+    super(initialState);
+
+    this.subscription = reducedState$.subscribe(this);
```
tlaundal, line 14: **This is indefinite subscription of the stream**

I believe this makes sense, as the global streams should always be active. The cost is, from what I can tell, only the size of `initialState`. We are still free to run initialization logic whenever we want.

anton164:
@tlaundal I changed the PR from draft status and added the `work-in-progress` label instead. Unfortunately PullPanda doesn't pick up PRs that are in draft mode since they are for some reason considered "not ready for review" although you request reviews 🤷 

Since our workflow heavily relies on pullpanda reminders I suspect no one reviewed it because pull panda didn't tell them to. Probably makes sense to avoid draft PRs unless we can find a setting to notify us about draft PR review requests.

holographic-principle:
Looks ok to me. I think BehaviorSubject is a great fit for our "global" streams.

src/createReducedStream.ts
anton164: I agree, would be nice to consolidate once we're able to :)

src/createReducedStream.ts
anton164: Does calling `.complete()` on the behavior tear down the subscription?

src/createReducedStream.ts
```diff
@@ -0,0 +1,54 @@
+import { BehaviorSubject, Observable, Subscription, Subject } from 'rxjs';
+import { ActionStream } from './types/helpers';
+import { RegisteredReducer, combineReducers } from './reducer';
+import { defaultErrorSubject } from './internal/defaultErrorSubject';
+import { markName } from './internal/markers';
+import { tag } from 'rxjs-spy/operators';
+
+export class ReducedStream<State> extends BehaviorSubject<State> {
```
anton164, line 8: From your PR description:
> The current implementation in this PR extends BehaviorSubject and subscribes the stream when created. This is not ideal, as next, error and complete are available as methods on the objects, though they should never be used. It also means the state can only be exposed through the property name value, which is very non-descriptive.

I think it's best if we can conclude on whether it's good or bad that `ReducedStream` is a behavior subject.

Perhaps it makes sense to rather implement our own stream which doesn't have a `.next()` method? I think it still makes sense to have `.complete()` so it could work with `takeUntil` rxjs operators.

I think the property name `value` is fine, since it's consistent with RxJS subjects.

src/createReducedStream.ts
chriskr: But why not named arguments? The order doesn't really reveal what the arguments are, except perhaps the name?

chriskr:
I'm missing a bit context here: what are the example use cases in our app where we plan to use this? The table view streams uses e.g. reduceState too, but this new stream can not be used there? We don't like to have permanent subscription for the view table stream?

src/createReducedStream.ts
tlaundal: This is the main reason this PR is WIP :)

I had some trouble formulating the logic when extending `Observable`, so I opted for extending `BehaviorSubject` just to get started. I'll look into `Observable` again.

src/createReducedStream.ts
tlaundal: I think that would be the best. I'm revisiting this implementation today.

src/createReducedStream.ts
anton164: 👍

tlaundal:
> I'm missing a bit context here: what are the example use cases in our app where we plan to use this? 

Me and @anton164 discovered that a lot of our streams don't really have any life cycle. They are created when the app starts, maybe when they are first needed, and then they never close. This allows the user to switch between different modules in the app without losing their work. This goes for more or less all our streams that are not view model streams, but also the view model streams for the search module. This tooling is initially focused at those streams.

src/createReducedStream.ts
tlaundal: I guess this is just a matter of opinion or preference. I think with well named variables it is very easy to understand something like:
```ts
const myState$ = createReducedStream(
  'myState$',
  initialState,
  reducers,
  action$
);
```
And that an object / named arguments just adds more overhead. Especially considering that the ordering will not be consistent.
```ts
const myState$ = createReducedStream({
  name: 'myState$',
  reducers,
  action$,
  seed: initialState,
});
```

tlaundal:
I think this is ready for merging and of course some later iteration to get it just right.

I changed the name of the function to `persistentReducedStream`, which follows our declarative naming scheme, and makes it a bit more clear that it will be subscribed on creation and not automatically unsubscribed. 

Sad to lose 0.4% of test coverage, but that's just the way it is.

src/createReducedStream.ts
chriskr: Yes, with well named variables it works.

PR #147, opened 2020-04-21
feat(markers): added marker for debounceTime

anton164:
- Ensured that a glitch is not detected over a debounceTime marker

Related PR in ardoq-front: https://github.com/ardoq/ardoq-front/pull/7055


src/internal/markers.ts
```diff
@@ -84,14 +85,19 @@ export type ZipMarker = MarkerBase<MarkerType.ZIP> & {
   readonly sources: (Marker | null)[];
 };
 
+export type DebounceTimeMarker = MarkerBase<MarkerType.DEBOUNCE_TIME> & {
+  readonly source: Marker | null;
```
tlaundal, line 89:
```suggestion
  readonly sources: [Marker | null];
```
The marker system was designed with the idea that each marker should use the same fields to list its sources and dependencies.
`detectGlitches` looks for the `sources` and `dependencies` fields to traverse the tree, and our streams-to-ardoq exporter also uses those fields to create the references.

src/operators/decorated.ts
```diff
@@ -40,3 +41,12 @@ export const startWith = ((...args: any) => (
   if (marker === null) return startWith$;
   return new MarkedObservable(startWith$, marker);
 }) as typeof operators.startWith;
+
+export const debounceTime: typeof operators.debounceTime = (
```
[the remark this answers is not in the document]
anton164: I think it's unnecessary to test the behaviour of debounceTime, but will add a test for the marker

anton164:
@tlaundal I've applied the changes we discussed (also updated the according PR in ardoq-front).

src/internal/markers.spec.ts
```diff
@@ -126,14 +123,34 @@ test('markZip includes sources', (t) => {
...
+  const piped$ = source$.pipe(markDebounceTime(source$, 0));
+
+  t.deepEqual(findMarker(piped$), {
+    type: MarkerType.DEBOUNCE_TIME,
+    sources: [TOP_MARKER],
+    time: 0,
+  });
+});
+
+test('debounceTime places marker with time', (t) => {
```
tlaundal, line 136: I think this test should be in `operators/decorated.spec.ts`

src/internal/markers.spec.ts
```diff
@@ -126,14 +123,34 @@ test('markZip includes sources', (t) => {
...
+
 const findMarkerOver: Macro<[OperatorFunction<any, any>]> = (t, operator) =>
   t.deepEqual(findMarker(source$.pipe(operator)), TOP_MARKER);
 findMarkerOver.title = (name) =>
   `findMarker finds marker over ${name} operator`;
 
 const coop = () => true;
 const emop = () => empty();
-test('debounceTime', findMarkerOver, debounceTime(0));
+// test('debounceTime', findMarkerOver, debounceTime(0));
```
tlaundal, line 153:
```suggestion
```

src/internal/markers.ts
```diff
@@ -84,14 +85,20 @@ export type ZipMarker = MarkerBase<MarkerType.ZIP> & {
   readonly sources: (Marker | null)[];
 };
 
+export type DebounceTimeMarker = MarkerBase<MarkerType.DEBOUNCE_TIME> & {
+  readonly sources: (Marker | null)[];
```
tlaundal, line 89:
```suggestion
  readonly sources: [(Marker | null)];
```
This change guarantees that once TS realizes the marker is a debounce marker, it will also know the array must have an element.

anton164: Aha, I missed that from your previous suggestion, my bad

PR #152, opened 2020-04-30
feat(reducers): added performance measurements to combineReducers

anton164:
If the window.performance API is available and a marker name is passed to
combineReducers it will add performance markers using the window.performance API.

The markers and measurements can be read and logged by interacting with
the performance API.

---
The benefit of using the window performance API is that they're available when using the browser for performance profiling:
Screenshot 2020-04-30 16 07 29
Screenshot 2020-04-30 16 07 38

I first tried to implement this in a way that performance measurements would also work when window isn't available by falling back to using the node performance API, but it wasn't straight forward. Since we don't have a use-case for that yet I think this is fine.

### DIFF
--- a/src/reducer.ts
+++ b/src/reducer.ts
@@ -1,5 +1,5 @@
 import { OperatorFunction, pipe, Observable, Subject } from 'rxjs';
-import { scan, map, filter } from 'rxjs/operators';
+import { scan, map, filter, tap } from 'rxjs/operators';
 import {
   VoidPayload,
   UnknownAction,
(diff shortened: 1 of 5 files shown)


src/reducer.ts
```diff
@@ -207,7 +211,8 @@ const ACTION_ORIGIN = Symbol('Action origin');
 export const combineReducers = <State>(
   seed: State,
   reducers: RegisteredReducer<State, any>[],
-  errorSubject: Subject<any> = defaultErrorSubject
+  errorSubject: Subject<any> = defaultErrorSubject,
+  performanceMarkerName?: string
```
anton164, line 215: I "lazily" added this as a last optional argument here, but I'm struggling to see a better place to put it. Any thoughts on this @tlaundal ?

tlaundal: Hmm, I think there isn't an obvious correct order here. Neither of them is more "important", and there is no guarantee that one of them will only be used together with the other.

Though, I hope this operator will turn more and more into an internal thing, so maybe the signature won't really matter that much?

The two alternatives I can think of:
 - A configuration object which takes either or both of them
 - Overloads and dynamic argument recognition, so the order doesn't matter. This should be possible as `Subject`/`object` and `string` are separate types

Of the two, I think the first one is the best, but I would also be fine with keeping it as it is.

anton164: thanks for the input 👍 

I agree and will refactor to a config object

PR #156, opened 2020-05-12
feat(state streams): updated persistentReducedStream

anton164:
- PersistentReducedStream no longer implements SubscriptionLike

- PersistentReducedStream must be explicitly started using .startReducing
to reduce values

  This ensures that less boilerplate is needed to set up state streams in an
  application and allows consumers to subscribe to reduced state streams
  before they are started.

  startReducing allows to recreate the inner subject of the state stream
  with a new initialState, which can be useful for hot-reloading

- PersistentReducedStream can be stopped using .stopReducing

BREAKING CHANGE: persistentReducedStream has a new interface

Relevant PR in ardoq-front: https://github.com/ardoq/ardoq-front/pull/7135
See the tests for more details on the behaviour.

### DIFF
--- a/src/persistentReducedStream.ts
+++ b/src/persistentReducedStream.ts
@@ -3,7 +3,6 @@ import {
   Observable,
   Subscription,
   Subject,
-  SubscriptionLike,
   Operator,
   Subscriber,
   TeardownLogic,
@@ -15,25 +14,85 @@ import { markName } from './internal/markers';
 import { tag } from 'rxjs-spy/operators';
 
 /**
- * A hot, persistent observable with a state field
+ * PersistentReducedStateStream is an Observable interface for state streams.
+ *
+ * The state stream has an inner subject which produces values to its subscribers.
+ *
+ * To begin reducing values, call `.startReducing(action$)`
+ * To stop reducing, call `.stopReducing()` (this closes the inner subject)
+ *
+ * The latest value of the state stream can be read synchronously by accessing
+ * `.state` (as long as the inner subject isn't closed)
  */
-export class StateStream<State> extends Observable<State>
-  implements SubscriptionLike {
+export class PersistentReducedStateStream<State> extends Observable<State> {
   private subject: BehaviorSubject<State>;
-  private subscription: Subscription;
+  private reducerSubscription?: Subscription;
+  public name: string;
+  private reducers: RegisteredReducer<State, any>[];
+  private errorSubject: Subject<any>;
 
-  constructor(initialState: State, reducedState$: Observable<State>) {
-    super();
-    this.subject = new BehaviorSubject(initialState);
-
+  /**
+   * Subscribes to the action stream and starts reducing this state stream.
+   * startReducing does not always immediately emit a value since the value
+   * might have already been emitted when the state stream was created.
+   *
+   * initialState can be passed to reset the state of the stream to some other value.
+   * This can be useful for hot-reloading
+   *
+   * startReducing will recreate the inner subject if it was previously closed
+   */
+  startReducing = (action$: ActionStream, initialState?: State) => {
+    // Reset the subject value if it doesn't match the desired initialState
+    // This is normally used for hot-reloading
+    if (this.subject.closed && initialState) {
+      this.subject = new BehaviorSubject(initialState);
+    } else if (initialState && this.subject.value !== initialState) {
+      this.subject.next(initialState);
+    }
     // NOTE: .subscribe(this.subject) does not work
     // All tests pass with it, but in practice there are cases
     // where the subject isn't updated as expected
-    this.subscription = reducedState$.subscribe({
-      complete: () => this.subject.complete(),
-      error: (err) => this.subject.error(err),
-      next: (state) => this.subject.next(state),
-    });
+    this.reducerSubscription = action$
+      .pipe(
+        combineReducers(initialState || this.subject.value, this.reducers, {
+          errorSubject: this.errorSubject,
+        }),
+        markName(this.name),
+        tag(this.name)
+      )
+      .subscribe({
+        complete: () => this.subject.complete(),
+        error: (err) => this.subject.error(err),
+        next: (state) => this.subject.next(state),
+      });
+    return this;
+  };
+
+  /**
+   * Stop reducing this state stream and unsubscribe from the action$.
+   *
+   * Marks the inner subject as closed to prevent new subscriptions
+   * (new subscriptions are allowed if the state stream is restarted)
+   * https://ncjamieson.com/closed-subjects/
+   */
+  stopReducing = () => {
+    if (this.reducerSubscription) {
+      this.reducerSubscription.unsubscribe();
+    }
+    this.subject.unsubscribe();
+  };
+
+  constructor(
+    name: string,
+    initialState: State,
+    reducers: RegisteredReducer<State, any>[],
+    errorSubject: Subject<any> = defaultErrorSubject
+  ) {
+    super();
+    this.name = name;
+    this.subject = new BehaviorSubject(initialState);
+    this.reducers = reducers;
+    this.errorSubject = errorSubject;
   }
 
   get state(): State {
@@ -48,11 +107,6 @@ export class StateStream<State> extends Observable<State>
     return this.subject.lift(operator);
   }
 
-  unsubscribe() {
-    this.subscription.unsubscribe();
-    this.subject.unsubscribe();
-  }
-
   _trySubscribe(subscriber: Subscriber<State>): TeardownLogic {
     return this.subject._trySubscribe(subscriber);
   }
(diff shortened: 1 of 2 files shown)


src/reducedStream.ts
```diff
@@ -0,0 +1,133 @@
...
+ * myState$.value === initialState // Will be true
+ *
+ * myState$.startReducing(action$) // To start reducing
+ * ```
+ *
+ * @param name The name of the stream, used for placing a marker and spy tag
+ * @param initialState The initial state of the stream
+ * @param reducers The reducers that build up the stream state
+ */
+export const reducedStream = <State>(
```
chriskr, line 126: If this is the main interface to create these kind of streams we should make it more clear what the purpose is. We cannot use these streams for e.g. view state streams, this is only meant for streams which should hold state over the whole life cycle of the application, view state streams on the other hand should start on subscription and stop when the according view is removed.

anton164: Agree, but I would like to wait with making that clarifications until we have proper tooling for the view state streams. For now RxBeach only provides tooling for creating these kinds of "permanent" state streams and IMO `reducedStream` is a good name for that.

I can update the RxBeach docs with some examples and clarifications.

src/reducedStream.ts
```diff
@@ -0,0 +1,133 @@
...
+  TeardownLogic,
+} from 'rxjs';
+import { ActionStream } from './types/helpers';
+import { RegisteredReducer, combineReducers } from './reducer';
+import { defaultErrorSubject } from './internal/defaultErrorSubject';
+import { markName } from './internal/markers';
+import { tag } from 'rxjs-spy/operators';
+
+export class ReducedStateStream<State> extends Observable<State>
+  implements SubscriptionLike {
```
chriskr, line 18: Looking at this again this is strange stuff. The Observable class is only used as an interface, wrapping the actual stream as a private subject member. The interface seems only partially implemented. It's difficult to really understand what this class tries to achieve. Also things are not really symmetric anymore, like subscribe and unsubscribe, subscribe does not start the actual subscription, but unsubscribe actually stops it.

anton164: I see your point, but I had to be "creative" in order to achieve the intended behaviour, as I was unable to find another way. Do you have suggestions for how this could be improved?

chriskr: I don't really understand the problem we are trying to solve here. Could you describe it?

src/reducedStream.ts
anton164: I updated the PR and renamed `reducedStream` to `persistentReducedStream`

src/reducedStream.ts
chriskr: The minimum would IMO be though to not use `unsubscribe` but rather `stopReducing` to keep it symmetric, otherwise this is doomed to cause confusion, I think.

anton164: Sure, we are trying to solve a set of problems related to creating persistent state streams in an application:

1. setting up & consuming state streams should be simple
    a. their values can be read synchronously (using `.state` like a behaviour subject)
    b. their values can be streamed (like any observable)
2. the streams are persistent, which means that we wish to subscribe to them and keep them "alive" through the application's lifecycle
3. they can be easily tested
4. they can be hot-reloaded during development

Let's look at each point separately. 

1. With the previous solution, I think we agreed that consuming & setting up streams was no longer simple, since the stream [had to be wrapped in get and setup methods](https://github.com/ardoq/ardoq-front/blob/aacb99716ce545fcd015ec96bfc6965bf191ca13/src/js/ardoq/branchId%24.ts#L16-L35). This PR fixes this by introducing a stream-like constructor "PersistentReducedStateStream". 
    - Consuming the stream is now straight-forward and simple. You can access the stream's value either by streaming it or by accessing `.state` synchronously. This interface is very similar to a behaviour subject (which inspired the solution).
    - Setting up the stream still requires [an additional step of calling startReducing](https://github.com/ardoq/ardoq-front/pull/7135/files#diff-ff0d9d26e235633b69491de7bc7b6cf5R31-R33), but this is unavoidable if we wish to maintain control of when persistent reduced streams are started.
2. Previously, we made streams persistent by `subscribing` and never removing this subscription. Now we do this by calling `startReducing` which emulates this setup step.
3. With this solution, the [state stream definition](https://github.com/ardoq/ardoq-front/pull/7135/files#diff-5dbdaf1d6e2303c79deffb56019a573d) file should ideally have no dependencies except pure actions & types. The action stream is injected using `startReducing` which makes it trivial to inject a mocked action stream. Unfortunately this is not the case for stream reducers yet, but this is difficult to solve while maintaining strict typing, so they might have to be mocked otherwise.
4. Finally, the proposed interface makes it trivial to solve hot-reloading, since we can read the state synchronously from the stream when it's hot-reloaded and [restart reducing of the stream with the latest value](https://github.com/ardoq/ardoq-front/blob/3b6e60d7b2c5d30f22057da923f668c8faf3a6ca/src/js/ardoq/startStateStreams.ts#L62-L90). 

Perhaps each point needs a bit more explanation & discussion, I'm happy to do that in a call :) From my point of view, it's very important to keep consumption of state streams simple and I think this PR achieves that. The internals of the stream could possibly be improved, but I think the idea of representing our "state streams" as custom observables similar to a behaviour subject makes sense. 

In conclusion, from the "ardoq-front" side, I think this implementation is the best we have, because:
- consuming streams is simple - you consume them like any other stream with the added benefit of having a `.state` accessor. 
- setting up streams  has the small caveat of calling `startReducing`, but I think we can accept this downside because it allows us to implement hot reloading & it makes testing simple

If you have any suggestions for how to improve the interface or think we are on the complete wrong path, that feedback is of course very welcome!

holographic-principle: Could we not achieve the same thing by simply extending BehaviorSubject, adding our custom "teardown" logic and possibly adding a hook to register the stream for "startReducing" (we could store them in a registry, that way we wouldn't have to import all of them in `startStateStreams.ts`, just that registry module)?

This is just by reading your explanation -- I haven't looked at the implementation in detail, although I agree with Krebs that the class is a bit weird right now.

anton164: @tlaundal 's initial PR extended BehaviorSubject, but we [landed upon extending Observable](https://github.com/ardoq/rxbeach/pull/142#discussion_r411507053) since it made more sense to expose state using `.state` and we didn't see any pros for extending behavior subject. I think one of main arguments was that the interface we've created is no longer a behavior subject, it's rather an interface to an internal behavior subject 🤓 😄 

Regarding a stream registry, would be nice with a concrete example that shows what you mean. I'm guessing that would require creating a static shared stream registry module among all the persistent reduced streams? Would we still be able to support hot reloading and controlling when reducing is started? Sounds like the approach could be a nice improvement if it removes the need for importing all of the streams in `startStateStreams` :)

anton164: Had a live discussion with @chriskr. I will look into the `unsubscribe` method and possibly split it into `stopReducing` and `unsubscribe` to make the interface more consistent & "symmetrical".

holographic-principle: Yes, "static" registry. How things would be registered is up for discussion, but I think it would "fix" startStateStreams. Not sure how it impacts hot reloading because I'm not that familiar with the hot reloading mechanism.

src/persistentReducedStream.ts
```diff
@@ -14,26 +13,68 @@ import { defaultErrorSubject } from './internal/defaultErrorSubject';
...
+      )
+      .subscribe({
+        complete: () => this.subject.complete(),
+        error: (err) => this.subject.error(err),
+        next: (state) => this.subject.next(state),
+      });
+    return this;
+  };
+
+  stopReducing = () => {
```
anton164, line 60: @chriskr this has now been updated with `stopReducing`, which is symmetrical to `startReducing`.

Consequently, `PersistentReducedStream` no longer implements the SubscriptionLike like a Subject normally would since we wish to have a `stopReducing` method instead of `unsubscribe` to "close" the state stream.

These tests document the updated behaviour:
https://github.com/ardoq/rxbeach/blob/8017283324cf4fdef9bf1194ec4e9b966f1bb5d7/src/persistentReducedStream.spec.ts#L93-L156

src/persistentReducedStream.ts
```diff
@@ -14,26 +13,68 @@ import { defaultErrorSubject } from './internal/defaultErrorSubject';
...
+   * Subscribes to the action stream and starts reducing this state stream.
+   * startReducing does not always immediately emit a value since the value
+   * might have already been emitted when the state stream was created.
+   *
+   * initialState can be passed to reset the state of the stream to some other value.
+   * This can be useful for hot-reloading
+   *
+   * startReducing will recreate the inner subject if it was previously closed
+   */
+  startReducing = (action$: ActionStream, initialState?: State) => {
```
anton164, line 33: `startReducing` also supports "restarting" of reducing if the inner subject is closed and `initialState` is passed, which is nice for hot-reloading: https://github.com/ardoq/ardoq-front/pull/7135/files#diff-ff0d9d26e235633b69491de7bc7b6cf5L27

Alternatively, we could split this into two methods `startReducing` and `restartReducing` to make it more explicit. Thoughts?

anton164:
Ready for a new review, the according ardoq-front PR has also been updated: https://github.com/ardoq/ardoq-front/pull/7135

src/persistentReducedStream.ts
```diff
@@ -14,26 +13,68 @@ import { defaultErrorSubject } from './internal/defaultErrorSubject';
...
+      .subscribe({
+        complete: () => this.subject.complete(),
+        error: (err) => this.subject.error(err),
+        next: (state) => this.subject.next(state),
+      });
+    return this;
+  };
+
+  stopReducing = () => {
+    this.subject.unsubscribe();
```
anton164, line 61: When reducing is stopped, the underlying `subject` will be closed. This means that reading the state of the state stream will throw an error (which is good because this might be outdated state).

chriskr, line 61: Why this and what is it doing? Is it needed?

anton164: Yes, we must unsubscribe from the inner subject to close it. You can read a bit more about closing subjects here: https://ncjamieson.com/closed-subjects/

In practice, I believe this is not strictly necessary, so we could also avoid it, but then we would also need to remove all the logic related to checking whether the inner subject is closed - we could instead check whether the `reducerSubscription` has been set up.

Furthermore, if we don't close the inner subject, reading the latest state of a "closed" reducer stream would still be possible, and in my opinion this shouldn't be allowed since it could lead to reading outdated state.

chriskr: Ok, makes sense. I think it would also make sense to add some of this info as a comment.  The key from that linked article in this context is IMO
> Well, the unsubscribe method in the Subject class doesn’t actually unsubscribe anything. Instead, it marks the subject as closed and sets its internal array subscribed observers — Subject extends Observable, remember — to null.

Didn't knew that.

anton164:
Updated the PR with documentation and squashed the commits. Should be ready for merge, but will do one more test with ardoq-front to ensure that everything works

src/persistentReducedStream.ts
```diff
@@ -15,25 +14,85 @@ import { markName } from './internal/markers';
...
+   * Stop reducing this state stream and unsubscribe from the action$.
+   *
+   * Marks the inner subject as closed to prevent new subscriptions
+   * (new subscriptions are allowed if the state stream is restarted)
+   * https://ncjamieson.com/closed-subjects/
+   */
+  stopReducing = () => {
+    this.subject.unsubscribe();
+    if (this.reducerSubscription) {
+      this.reducerSubscription.unsubscribe();
```
tlaundal, line 81: Should this go before `subject.unsubscribe()`?
I have a feeling it won't matter in practice, but theoretically couldn't the `action$` get a new value between these two lines, leading to the `reducerSubscription` trying to `next`  value to the `subject`?

PR #162, opened 2020-05-29
Add a stream registry

tlaundal:
The stream registry does two things:
 - Prevents stream with duplicate names
 - Provides a central place to start all streams

This removes boilerplate code from users of rxbeach (no need for a stream registry in the application), and ensures created streams will always be started.

One caveat with this, which shouldn't really be an issue, is that the module defining a stream with `persistentReducedStream` must be imported somewhere in the application. This will usually happen naturally, and does not seem to be an issue in ardoq-front.

Fixes ARD-5295


src/persistentReducedStream.spec.ts
```diff
@@ -203,3 +206,10 @@ test(
     state$.startReducing(action$);
   })
 );
+
+test('persistentReducedStream should throw exception when accessing state after stopped reducing', (t) => {
```
anton164, line 210: 👍

src/stateStreamRegistry.ts
```diff
@@ -0,0 +1,87 @@
...
+      stream.stopReducing();
+    }
+  }
+
+  /**
+   * Get the latest state of each stream, for input into `startReducing`
+   *
+   * If `stopReducing` has been called before this, an exception will be thrown.
+   */
+  getStates(): Map<string, any> {
```
anton164, line 80: Nice! It's great to have this kind of interface for debugging as well :)

src/stateStreamRegistry.ts
```diff
@@ -0,0 +1,87 @@
...
+   * thrown.
+   *
+   * If the registry has been started, the registered stream will also be
+   * started.
+   *
+   * @param stream The stream to register
+   */
+  register(stream: PersistentReducedStateStream<any>) {
+    if (this.streams.has(stream.name)) {
+      throw new Error(`Duplicate stream name: ${stream.name}`);
```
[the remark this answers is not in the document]
anton164: From testing a bit locally, this is actually a bit tricky to achieve 🤔 

When `branchId$` is hot-reloaded, the only way to "dispose of it" is from within `branchId$` so it means that every state stream file would need unregistering logic (not ideal).

One alternative is to handle `re-registering` of a stream as hot-reloading, but this can lead to problems if you inadvertently register two streams with the same name.

I.e.:

```typescript
  register(stream: PersistentReducedStateStream<any>) {
    if (this.started !== null) {
      const duplicateStream = this.streams.get(stream.name);
      if (duplicateStream) {
        duplicateStream.stopReducing();
      }
      stream.startReducing(
        this.started.action$,
        this.started.states.get(stream.name)
      );
    }
    this.streams.set(stream.name, stream);
  }
```

^ this doesn't feel right, and it will cause problems if you have duplicate stream names, but I think it solves the hot-reloading issue :P

src/PersistentReducedStateStream.ts
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * initialState can be passed to reset the state of the stream to some other value.
+   * This can be useful for hot-reloading
+   *
+   * startReducing will recreate the inner subject if it was previously closed
+   */
+  startReducing = (action$: ActionStream, initialState?: State) => {
+    // Reset the subject value if it doesn't match the desired initialState
+    // This is normally used for hot-reloading
+    if (this.subject.closed && initialState) {
```
anton164, line 47: Just found a small (unrelated) bug. This will not hot-reload correctly if `initialState` is falsy.

It should be:
```typescript
if (this.subject.closed && initialState !== undefined) {
```

src/stateStreamRegistry.ts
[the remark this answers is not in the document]
anton164: After some thinking, an alternative solution for the duplicate problem could be to only throw an error if the registry is not started yet. 

Then it becomes:
```typescript
  register(stream: PersistentReducedStateStream<any>) {
    const previouslyRegisteredStream = this.streams.get(stream.name);
    if (this.started !== null) {
      let state = this.started.states.get(stream.name);
      // Since we've already started the stream registry
      // assume that we are "re-registering" the stream due to hot-reloading
      if (previouslyRegisteredStream) {
        state = previouslyRegisteredStream.state;
        previouslyRegisteredStream.stopReducing();
      }
      stream.startReducing(this.started.action$, state);
    } else if (previouslyRegisteredStream) {
       throw new Error(`Duplicate stream registered before starting the stream registry. name: ${stream.name}`);
    }
    this.streams.set(stream.name, stream);
  }
```

This seems to solve the hot-reloading problem and the duplicate stream problem if my assumptions are correct :)

src/PersistentReducedStateStream.ts
tlaundal: Good catch!
I try to stick to `===` and `!==`, but didn't review this properly when copying over.

src/stateStreamRegistry.ts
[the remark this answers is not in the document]
tlaundal: Thanks for looking into this. I didn't have the chance to test hot reloading before publishing :/

I like the idea of allowing re-registration only after the registry has started, but I also don't quite like the state dependent behavior.
Another approach could be to handle this differently depending on dev/prod environment. So in prod we would always throw an error, while in dev we would always re-register, perhaps with a `console.warn`.
`tsdx` (which we are not, yet, using) has proper support for [separate dev and prod bundles](https://github.com/jaredpalmer/tsdx#development-only-expressions--treeshaking), so I think this might be a viable approach.

src/PersistentReducedStateStream.ts
```diff
@@ -0,0 +1,120 @@
...
+ * To begin reducing values, call `.startReducing(action$)`
+ * To stop reducing, call `.stopReducing()` (this closes the inner subject)
+ *
+ * The latest value of the state stream can be read synchronously by accessing
+ * `.state` (as long as the inner subject isn't closed)
+ */
+export class PersistentReducedStateStream<State> extends Observable<State> {
+  private subject: BehaviorSubject<State>;
+  private reducerSubscription?: Subscription;
+  public name: string;
```
holographic-principle, line 30: Perhaps members should be sorted in private, protected, public order.

src/PersistentReducedStateStream.ts
holographic-principle, line 47: This whole block of logic could be a method with a descriptive name.

src/PersistentReducedStateStream.ts
```diff
@@ -0,0 +1,120 @@
...
+      this.subject = new BehaviorSubject(initialState);
+    } else if (initialState && this.subject.value !== initialState) {
+      this.subject.next(initialState);
+    }
+    // NOTE: .subscribe(this.subject) does not work
+    // All tests pass with it, but in practice there are cases
+    // where the subject isn't updated as expected
+    this.reducerSubscription = action$
+      .pipe(
+        combineReducers(initialState || this.subject.value, this.reducers, {
```
holographic-principle, line 57: As per @anton164 's comment, `&&` and `||` logic related to `initialState` should be cleaned up to assume possible falsey values.

src/PersistentReducedStateStream.ts
```diff
@@ -0,0 +1,120 @@
...
+   * Subscribes to the action stream and starts reducing this state stream.
+   * startReducing does not always immediately emit a value since the value
+   * might have already been emitted when the state stream was created.
+   *
+   * initialState can be passed to reset the state of the stream to some other value.
+   * This can be useful for hot-reloading
+   *
+   * startReducing will recreate the inner subject if it was previously closed
+   */
+  startReducing = (action$: ActionStream, initialState?: State) => {
```
holographic-principle, line 44: Should hot reloading actually reset the state? Is this desirable from a dev UX perspective?

anton164: So far our goal has been for hot-reloading to keep the state (similar to how react-hot-loader works).

holographic-principle: Right, that's how I would prefer it too. I'm just wondering if there's any reason to pass initialState to this function.

anton164: Aha, makes sense.

Having an argument for `initalState` is what allows us to avoid resetting the state when hot-reloading. When hot-reloading, the code that sets up the stream is re-ran, so if we didn't have a way to pass the "initial state", we wouldn't be able to insert the state that was built up before hot reloading (the stream would simply be reset to it's default state).

Perhaps the naming could be better, but the intention here is that `initialState` allows you to override what you use for the initial state of the stream. Normally the `initalState` is set to `defaultState`, however in the hot-reloading case `initialState` can be overriden to ensure that the stream state does not reset 😛  It's a bit of a mouthful.

Perhaps this code snippet demonstrates it better:

```typescript
// startStateStreams.ts
import action$ from 'action$';
import stateStream$ from 'somewhere';


// When hot-reloading module.hot.data.lastestStreamState will be set
// because the dispose handler is ran when this module was disposed
stateStream$.startReducing(action$, module.hot?.data?.latestStreamState);

if (module.hot) {
  module.hot.dispose(data => {
    data.latestStreamState = stateStream$.state;
    stateStream$.stopReducing();
  });
}
```

holographic-principle: Yeah, but don't we have `this.subject.value`? Maybe I am missing the point here :see_no_evil:

anton164: `this.subject.value` can't be accessed when the stream has stopped reducing because it's closed (https://ncjamieson.com/closed-subjects/)

holographic-principle: Ok, got it now. Thanks.

tlaundal:
I have fixed the comments.
Went for the `NODE_ENV` based approach for duplicate names. It is as strict as possible, while still allowing hot reloading of the streams. 

Tested and works together with ardoq/ardoq-front#7237

src/persistentReducedStream.spec.ts
```diff
@@ -155,6 +158,36 @@ test(
...
+
+  t.deepEqual(state$.state, state);
+};
+restartWithState.title = (state) =>
+  `persistentReducedStream should allow restarting with specified state ${state}`;
+
+test('number', restartWithState, 12);
+test('zero', restartWithState, 0);
+test('false', restartWithState, false);
+test('null', restartWithState, null);
```
anton164, line 189: 💯

src/persistentReducedStateStream.ts
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * initialState can be passed to reset the state of the stream to some other value.
+   * This can be useful for hot-reloading
+   *
+   * startReducing will recreate the inner subject if it was previously closed
+   */
+  startReducing = (action$: ActionStream, initialState?: State) => {
+    // Reset the subject value if it doesn't match the desired initialState
+    // This is normally used for hot-reloading
+    if (initialState !== undefined) {
```
holographic-principle, line 48: I would still extract this block into a method.

PR #164, opened 2020-06-04
fix: upgraded vulnerable dependency

RenaudF:
https://ardoqcom.atlassian.net/browse/ARD-5317

I got there by:
 - adding `"mkdirp": "0.5.4"` to `package.json`
 - running `yarn install`
 - manually replacing the `mkdirp` version from `0.5.1` to `0.5.4` in `yarn.lock`
 - removing `"mkdirp": "0.5.4"` from `package.json`
 - running `yarn install`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3151,15 +3151,15 @@ minimist-options@^3.0.1:
     arrify "^1.0.1"
     is-plain-obj "^1.1.0"
 
-minimist@0.0.8:
-  version "0.0.8"
-  resolved "https://registry.yarnpkg.com/minimist/-/minimist-0.0.8.tgz#857fcabfc3397d2625b8228262e86aa7a011b05d"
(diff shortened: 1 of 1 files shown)


RenaudF:
> This branch brings it down to 18, which is a start, but doesn't really fix the issue. We still have multiple dependency paths to `minimist` left.

That's not an issue, the vulnerability is only for `minimist` versions below `0.2.1`, see:
Screenshot from 2020-06-04 16-03-22 

> Also, is there a risk for yarn overwriting this fix when we update other dependencies, because it only exists in `yarn.lock`?

If you run `yarn install` or `yarn install -f` again, the `yarn.lock` doesn't change.

RenaudF:
There is still a version of `minimist` that's too low, but that comes from `"standard-version": "8.0.0"`

PR #233, opened 2020-07-30
Support namespace in persistentReducedStreams

tlaundal:
Adds rudimentary namespace support to persistentReducedStreams. 

Review commit by commit, and take a look at the commit messages.

### DIFF
--- a/src/persistentReducedStream.ts
+++ b/src/persistentReducedStream.ts
@@ -1,8 +1,12 @@
 import { Subject } from 'rxjs';
 import { RegisteredReducer } from './reducer';
-import { defaultErrorSubject } from './internal/defaultErrorSubject';
 import { PersistentReducedStateStream } from './persistentReducedStateStream';
 import { stateStreamRegistry } from './stateStreamRegistry';
+
+type PersistentReducedStreamOptions = {
+  errorSubject?: Subject<any>;
+  namespace?: string;
+};
 
 /**
  * Creates and registers a persistent reduced state stream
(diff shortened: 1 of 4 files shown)


src/namespace.ts
```diff
@@ -59,4 +59,17 @@ export const namespaceActionDispatcher = (
...
+  dispatchNamespaced(namespace, parentDispatcher, action);
+
+/**
+ * Add the namespace to the action and dispatch it.
+ *
+ * @param namespace Namespace to dispatch action with
+ * @param parentDispatcher Action dispatcher to dispatch action with
+ * @param action The action to namespace and dispatch
+ */
+export const dispatchNamespaced = (
```
chriskr, line 71: I think the interface is too complicated to use namespaces. Ideally the namespace should IMO just be an optional argument of `dispatchAction`, basically `dispatchAction(action: Action, namespace?: Namespace)` with `Namespace` being an enum. Accordingly in the creating stream functions the namespace would be the last optional argument too. By default the namespace is null.

tlaundal: I agree, this is too complicated. I looked into something like an optional argument to `dispatchAction`, but I don't remember why I didn't go for it.

I will take a look at it again, and probably create a PR which brings the `action$` and `dispatchAction` into RxBeach.

tlaundal: New PR is up in #234

src/persistentReducedStream.ts
```diff
@@ -39,13 +43,14 @@ export const persistentReducedStream = <State>(
   name: string,
   initialState: State,
   reducers: RegisteredReducer<State, any>[],
-  errorSubject: Subject<any> = defaultErrorSubject
+  { errorSubject, namespace }: PersistentReducedStreamOptions = {}
```
tlaundal, line 46: It's a bit unfortunate that we have to introduce an option object here.
In a while, I think we should reconsider how much we have used `errorSubject` and consider converting back to normal arguments with `errorSubject` as the last, optional, one.

chriskr: Could you not use two signatures, one with the errorSubject and one without? rxjs does that a lot?

tlaundal: I don't think that's worth it at this point. I'd like to delay introducing complexity like that as far as possible.
I expect that in a few months, the signature will end up as `..., namespace?: string, errorSubject?: Subject`, as we don't really use the `errorSubject` at all, it's just there for completeness sake

PR #234, opened 2020-07-30
feat: introduce the action$ in RxBeach

tlaundal:


tlaundal:
Please take a quick look at this even though I merged already, @chriskr

src/action$.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param action The action to dispatch to action$
+ * @param namespace Optional namespace to add to the action
+ */
+export const dispatchAction: ActionDispatcher = (action, namespace) => {
+  if (namespace === undefined) {
+    actionSubject$.next(action);
+  } else {
+    actionSubject$.next(_namespaceAction(namespace, action));
+  }
+};
```
chriskr, line 33: Yes, this looks good

src/types/helpers.ts
```diff
@@ -4,7 +4,10 @@ import { UnknownAction } from '../internal/types';
 
 export type ActionStream = Observable<UnknownAction>;
 
-export type ActionDispatcher = (action: UnknownAction) => void;
+export type ActionDispatcher = (
+  action: UnknownAction,
+  namespace?: string
```
chriskr, line 9: Should we not use well defined namespaces, e.g use an `enum Namespace` ?

tlaundal: `string` allows us to pass string enums. This means we can keep RxBeach separate from ardoq-front, with no need for it to know the namespaces we use in ardoq-front. In ardoq-front we can define a string enum if we want.

PR #241, opened 2020-08-10
feat: forward namespaces to reducers

tlaundal:
The motivation for this change is to allow more reducers to be statically created. See [this example](https://github.com/ardoq/ardoq-front/blob/master/src/js/ardoq/search/QueryEditor/queryEditor%24.ts#L195-L212) from ardoq-front.

As you can see from the tests. The syntax is not ideal for void actions (with the underscores/unused arguments).

### DIFF
--- a/src/persistentReducedStream.spec.ts
+++ b/src/persistentReducedStream.spec.ts
@@ -4,6 +4,7 @@ import { of, Subject, empty } from 'rxjs';
 import { marbles } from 'rxjs-marbles/ava';
 import { incrementMocks } from './internal/testing/mock';
 import { map } from 'rxjs/operators';
+import { reducer } from './reducer';
 
 const { reducers, actionCreators, handlers } = incrementMocks;
 const { actions, numbers, errors } = incrementMocks.marbles;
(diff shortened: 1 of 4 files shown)


src/reducer.spec.ts
```diff
@@ -121,3 +121,29 @@ test(
...
+
+test(
+  'combineReducers should forward namespace',
+  marbles((m) => {
+    const action$ = m.hot('  1', actions);
+    const expected$ = m.hot('2', numbers);
+
+    const verifyNamespaceReducer = reducer(
+      actionCreators.incrementOne,
+      (_, __, namespace) => {
```
tlaundal, line 133: This is an example of the unfortunate syntax. This is how we will use it for reset actions in ardoq-front.

src/persistentReducedStream.spec.ts
```diff
@@ -277,3 +278,32 @@ test(
...
+test(
+  'persistentReducedStream forwards namespace to reducers',
+  marbles((m) => {
+    const action$ = m.hot('  --n-', actions);
+    const expected$ = m.hot('1-2-', numbers);
+    const initialState = 1;
+
+    const verifyNamespaceReducer = reducer(
+      actionCreators.incrementOne,
+      (state, _, namespace) => {
```
andreajl, line 291: In the action, the namespace is within the `meta` object. This can be done for the reducer as well, or what do you think? It might make it easier to extend this in the future if more variables are passed down (Not sure if that is planned tho). I'm fine with this if you think this is the last 👍

chriskr: If there will be more then I think we need to start to name the arguments.

src/reducer.ts
```diff
@@ -19,7 +19,8 @@ const wrapInArray = <T>(val: T | T[]): T[] =>
 
 export type Reducer<State, Payload = VoidPayload> = (
   previousState: State,
-  payload: Payload
+  payload: Payload,
+  namespace?: string
```
chriskr, line 23: I still think default namespace being null is better than optional.

tlaundal: Yes. I've created an issue for that ARD-6151
However, it is out of scope for this change

src/persistentReducedStream.spec.ts
tlaundal: I don't plan on passing along more variables currently. As long as we are in `0.x.x`, and are unafraid of breaking changes and refactoring, I think we should keep it like this for now, but it might make sense at some point to pass the whole action or meta object.

tlaundal:
> If I get this right, using a namespace in the reducer is more an exception? The prefer pattern would be to use `withNamespace` to filter the action stream?

Yes, this will rarely be used, but we have one use case in ardoq-front. Consider a reset action that has different defaults for different streams - then the reducer would need to know what stream it is currently reducing. So this is actually orthagonal to `withNamespace`.

src/reducer.ts
chriskr: Sure, that was more a note

PR #258, opened 2020-09-15
chore(deps-dev): update @ardoq/shared-configs to 0.1.2

tlaundal:


tlaundal:
Needs the new rules from ardoq/shared-configs#6

PR #272, opened 2020-10-08
docs: remove duplicate examples and add missing parentheses

OyvindSabo:


tlaundal (approved):
:100:
Thanks for removing the duplicates as well

PR #324, opened 2021-06-24
Enhance useStream with default value

michaltaberski:


tlaundal (changes requested):
Nice idea, could you give a use case example? I'm a bit curious as to how the call site would look with and without this.

Requesting changes because the docstring must be updated and we should add a small test for this.

tlaundal (approved):
Vacation review on mobile, but LGTM!

michaltaberski:
It's not critical PR, so it can wait for a proper review once you are back. Enjoy!

src/react/connect.ts
```diff
@@ -13,9 +13,15 @@ export type NOT_YET_EMITTED = typeof NOT_YET_EMITTED;
  * has no guarantee for when the first emit will happen.
```
tlaundal, line 13:
```suggestion
 * has no guarantee for when the first emit will happen. This initial
 * value may be overridden with the `defaultValue` argument.
```

src/react/connect.ts
```diff
@@ -13,9 +13,15 @@ export type NOT_YET_EMITTED = typeof NOT_YET_EMITTED;
...
  *
  * @param source$ Stream that provides the needed values
+ * @param defaultValue Default value returned on init until stream emits new value
  */
-export const useStream = <T>(source$: Observable<T>): T | NOT_YET_EMITTED => {
-  const [value, setValue] = useState<T | NOT_YET_EMITTED>(NOT_YET_EMITTED);
+export const useStream = <T>(
+  source$: Observable<T>,
+  defaultState?: T
```
tlaundal, line 20:
```suggestion
  defaultValue?: T
```
Keep in line with the `@param` above.

tlaundal (approved):
:100:

PR #352, opened 2021-09-08
fix: ARD-9960 fixed carry bug

tomoboy:
The carry operator had a bug that caused it to subscribe twice to the routines where it was used. More detailed description. 
TLDR: caused by difference between hot vs cold observables

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 dist
 .nyc_output
 coverage
+.idea
--- a/src/operators/operators.spec.ts
+++ b/src/operators/operators.spec.ts
@@ -223,7 +223,19 @@ test(
     ).toBeObservable(expected);
   })
 );
-
+test(
+  'carry only subscribes to the original observable once',
+  marbles((m) => {
+    const source$ = m.cold(' f', { f });
+    const expected = m.hot('1', {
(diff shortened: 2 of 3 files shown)


chriskr (approved):
LGTM

It's still subscribed twice, but with a cold observable side effects like showing a dialog are done twice because with a cold observable each subscriber gets its dedicated observable.

andreajl (approved):
Amazed that we have not seen this before 🧙‍♀️

tlaundal (commented):
Good fix!

I have three comments, though:
 - I think `share` is the appropriate operator for making the observable hot in this case, as we shouldn't need the replay (both subscriptions happen at the same time)
 - The function documentation should mention it makes the observable/stream hot
 - Could we write some small tests to verify this? With marbles we could check that the original observable is only subscribed once, and perhaps we could check that some side effect upstream only happens once as well (eg. `obs.pipe(tap(d => sideEffect(d)), carry(d => ...))`)

tomoboy:
@tlaundal Not 100% sure I wrote the test in an optimal way, but at least it passes for the new `carry` implementation and fails for the old one.

tlaundal (approved):
Nice!

Feel free to hit the merge button, but I think we can wait to cut a release until action name changes are ready as well.

PR #478, opened 2022-08-16
Upgrade rxbeach to rxjs7

tlaundal:
Upgrade all dependencies to latest version, and in particular update RxJS to 7.5.6.

RxJS has a bunch of breaking changes, the consequences of these in RxBeach are:
- `merge` and `combineLatest` exported from the operators namespace have been renamed with the `with` suffix
- Operators reexported from RxJS in RxBeach will have changed signatures as in RxJS
- The class `PersistentReducedStateStream` has been deleted, as it relied on extending `Observable`, and would probably not work in RxJS 8
- `persistentReducedStream` now returns instances of the new class `ObservableState`, which does not extend `Observable`. Some types will need to be changed.
- As `ObservableState` implements `InteropObservable` it integrates nicely into RxJS, but some custom tooling might need to be updated to accept `ObservableInput` as arguments, and use the `from` function from RxJS

This PR is read for review, and possibly merge. It is in use in a team environment in ardoq-front for testing by QA. Once this PR passes review, we need to:
 - Drop the prerelease commit
 - Merge to master (should be fast-forward)
 - Release version 1.0.0


src/observableState.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @deprecated Only for use by RxJS
+   */
+  [Symbol.observable](): Subscribable<State> {
+    return {
+      subscribe: this.subscribe,
+    };
+  }
+
+  /**
+   *
```
tlaundal, line 58: TODO

src/observableState.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.subscription = this.source$.subscribe(this.subject);
+
+    // NOTE: .subscribe(this.subject) does not work
+    // All tests pass with it, but in practice there are cases
+    // where the subject isn't updated as expected
+    // .subscribe({
+    //   complete: () => this.subject.complete(),
+    //   error: (err) => this.subject.error(err),
+    //   next: (state) => this.subject.next(state),
+    // });
```
tlaundal, line 73: Seems this can be removed, unless QA has some input

src/operators/decorated.ts
```diff
@@ -23,19 +23,24 @@ export const withLatestFrom = ((...dependencies: Observable<unknown>[]) =>
...
+export const combineLatestWith = ((
+    source: Observable<unknown>,
+    ...sources: Observable<unknown>[]
+  ) =>
   (observable$: Observable<unknown>) =>
     observable$.pipe(
-      operators.combineLatest(...sources),
-      markCombineLatest([observable$, ...sources])
-    )) as typeof operators.combineLatest;
+      operators.combineLatestWith([source, ...sources]),
```
tlaundal, line 32: Did we make a mistake here?

tlaundal: We did!
```suggestion
      operators.combineLatestWith(source, ...sources),
```

src/reducer.ts
```diff
@@ -162,14 +172,14 @@ type ReducerCreator = {
...
 
 export const reducer: ReducerCreator = <State>(
-  trigger: UnknownActionCreator | UnknownActionCreator[] | Observable<any>,
+  trigger: UnknownActionCreator | UnknownActionCreator[] | ObservableInput<any>,
   reducerFn: Reducer<State, any>
 ) => {
   const wrapper = (state: State, payload: any, namespace?: string) =>
     reducerFn(state, payload, namespace);
-  if (trigger instanceof Observable) {
+  if (!Array.isArray(trigger) && isObservableInput(trigger)) {
```
tlaundal, line 180: This doesn't really sit well with me, but I think we can fix it after this release

tlaundal (commented):
- Had to swap `nyc` for `c8`
- `persistentReducedStream` now returns an `ObservableState`, which does not extend `Observable`
- Changes from RxJS 7

src/isObservableInput.ts
```diff
@@ -0,0 +1,12 @@
+import { ObservableInput, from } from 'rxjs';
+
+export const isObservableInput = (
```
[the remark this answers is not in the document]
tlaundal: Used internally, see comment below

rafaa2:
Should we change it to ready to review?

tlaundal:
> @tlaundal Should we change it to ready to review?

Not yet. I'm making some small changes and plan to rewrite the history. We need to rewrite the history to get a proper changelog when we run the next release.

davidnegley (approved):

src/observableState.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.subject.unsubscribe();
+  };
+
+  get state(): State {
+    return this.subject.getValue();
+  }
+  get destroyed(): boolean {
+    return this.subject.closed;
+  }
+}
```
chriskr, line 104: Looks nice! Quite a bit simpler than the old version.

src/reducer.ts
chriskr: As we discussed, I think from an interface and readability point of view it would be better to have a `actionReducer` and `stateReducer`, that would probably also allow to make the code here a bit simpler.

chriskr (approved):
LGTM (not tested)

PR #479, opened 2022-08-17
feat!: remove all marker functionality

tlaundal:
The marker functionality was introduced to support building a tree of
streams, in effect determining how different observables relate to
eachother without needing to watch values pass through. This was never
used to its full potential. The effect on the codebase of RxBeach and
all users is quite big, as it means we need to reexport decorated
versions of operators from RxJS.

Removing this functionality reduces the size of the library and hopefully makes it more maintainable.

BREAKING CHANGE: RxBeach no longer re-exports operators from RxJS.
These operators must now be imported directly from RxJS.


tlaundal:
Blocked by #478

rafaa2 (approved):
LGTM

PR #484, opened 2022-08-25
Refresh routines

rafaa2:


src/internal/routineFunc.ts
```diff
@@ -121,3 +122,8 @@ export interface RoutineFunc {
     ...fns: OperatorFunction<any, any>[]
   ): Routine<unknown>;
 }
+
+export type SubscribeRoutineFunc = (
```
tlaundal, line 126: Is there a specific reason this should be a separate type, and not just be in-lined in the method in `routinesRegistry`?

src/routinesRegistry.ts
```diff
@@ -0,0 +1,110 @@
...
+   * when they are subscribed. If a routine dispatches an action when it is
+   * subscribed, only some of the routines will receive the action.
+   *
+   * @param routineToSubscribe The routine to subscribe
+   * @param config Containes the errorSubject and action stream.
+   *  both errorSubject and action$ have a default value.
+   * @returns A Subscription object
+   * @see defaultErrorSubject
+   */
+  subscribeRoutine: SubscribeRoutineFunc = (
```
tlaundal, line 91: I think this should be made into a standalone function next to this class (in the same module, but not exported).
Also, since it will only be used by this class, I think it's better to pass all three args as normal args, instead of the object.
IMO the `errorSubject` should be provided to the registry  in the same way as the `actino$`.

tlaundal, line 91: I think this should be made into a standalone function next to this class (in the same module, but not exported).
Also, since it will only be used by this class, I think it's better to pass all three args as normal args, instead of the object.
IMO the `errorSubject` should be provided to the registry  in the same way as the `action$`.

tlaundal, line 91: I think this should be made into a standalone function next to this class (in the same module, but not exported).
Also, since it will only be used by this class, I think it's better to pass all three args as normal args, instead of the object.
IMO the `errorSubject` should be provided to the registry  in the same way as the `action$`.

src/routinesRegistry.ts
```diff
@@ -0,0 +1,110 @@
...
+    const errorSubject: Subject<any> = e ?? defaultErrorSubject;
+    const action$: ActionStream = a$ ?? this.action$;
+
+    return action$
+      .pipe(
+        routineToSubscribe,
+        catchError((err, stream) => {
+          errorSubject.next(err);
+          return stream;
+        })
```
tlaundal, line 104: An unrelated idea: It would be nice to have a `forwardErrorAndContinue` operator that abstracts away this:
```ts
action$.pipe(
    routineToSubscribe,
    forwardErrorAndContinue(errorSubject)
)
```

tlaundal (commented):
I think this looks good! Would like to have a new review when everything, including the TODO and tests are ready, though.

One note that didn't fit in the review comments:
The `Routine` type in `internal/routineFunc` has "See `collectRoutines` for documentation" as docstring, I we should probably update this.

PR #497, opened 2022-09-02
feat(react): add `connectInstance` function

tlaundal:
The new `connectInstance` function works the same as `connect`, but
takes a function that takes an instance name and produces
`ObservableInput`. This is well suited for namespaced streams.

### DIFF
--- a/src/react/connect.ts
+++ b/src/react/connect.ts
@@ -45,16 +45,53 @@ export const useStream = <T>(
  * provided by the stream. In these cases the values from the stream will
  * override any values passed as props.
  *
- * @param WrappedComponent Component that will recieve props from the stream
+ * @param Component Component that will recieve props from the stream
  * @param stream$ Stream that will provide props to the component
- * @see useStream
+ * @returns A component that renders the original component with props retrieved
+ *  from the stream
+ * @see {@link useStream}
  */
 export const connect =
-  <Props extends {}, Observed extends Partial<Props>>( // eslint-disable-line @typescript-eslint/ban-types
+  <Props extends Record<string, unknown>, Observed extends Partial<Props>>(
     Component: ComponentType<Props>,
     stream$: ObservableInput<Observed>
   ) =>
   (props: Omit<Props, keyof Observed>) => {
+    const value = useStream(stream$);
+
+    if (value === NOT_YET_EMITTED) return null;
+
+    // Typescript doesn't recognize this as Observed & Props for some reason
+    // Question on StackOverflow: https://stackoverflow.com/q/60758084/1104307
+    const newProps = { ...props, ...value } as Props & Observed;
+
+    return React.createElement(Component, newProps);
+  };
+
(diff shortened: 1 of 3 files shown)


src/react/connect.ts
```diff
@@ -45,12 +45,14 @@ export const useStream = <T>(
...
+ * @param Component Component that will recieve props from the stream
  * @param stream$ Stream that will provide props to the component
- * @see useStream
+ * @returns A component that renders the original component with props retrieved
+ *  from the stream
+ * @see {@link useStream}
  */
 export const connect =
-  <Props extends {}, Observed extends Partial<Props>>( // eslint-disable-line @typescript-eslint/ban-types
+  <Props extends Record<string, unknown>, Observed extends Partial<Props>>(
```
rafaa2, line 55: out of curiousity, why?

src/react/connect.ts
```diff
@@ -65,3 +67,38 @@ export const connect =
...
+ * component.
+ *
+ * @param Component Component that will receive props from the stream
+ * @param createInstanceStream Function to create the stream for each instance
+ *  of the component
+ * @returns A component that renders the original component with props retrieved
+ *  from the stream
+ * @see {@link connect}
+ */
+export const connectInstance =
```
rafaa2, line 88: I am curious how to use this with a stream

src/react/connect.ts
tlaundal: This will restrict `Props` to be objects (so you can no longer pass primitives like a number or a boolean), and it means we don't need to have the lint ignore line there.

src/react/connect.ts
tlaundal:
```ts
connectInstance(MyComponent, (viewInstanceId) => action$.pipe(withNamespace(viewInstanceId), ...))
```
The important part is that it accepts a function that receives a view instance name / id.

src/react/connect.ts
```diff
@@ -65,3 +67,38 @@ export const connect =
...
+  ) =>
+  (props: Omit<Props, keyof Observed>) => {
+    const [stream$] = useState(() => createInstanceStream(nextInstanceName()));
+    const value = useStream(stream$);
+
+    if (value === NOT_YET_EMITTED) return null;
+
+    // Typescript doesn't recognize this as Observed & Props for some reason
+    // Question on StackOverflow: https://stackoverflow.com/q/60758084/1104307
+    const newProps = { ...props, ...value } as Props & Observed;
```
holographic-principle, line 101: Does it think `value` still might be `NOT_YET_EMITTED`? Does casting `value` to `Observed` help?

holographic-principle: BTW, I would not use the same name for the type and the constant in the case of `NOT_YET_EMITTED`...

holographic-principle (approved):

rafaa2 (approved):

PR #520, opened 2022-09-23
Feat: add reduced stream

rafaa2:
We currently have an operator called reduceState, which should probably be a separate function called reducedStream to show the relation and difference from persistentReducedStream.

In ardoq-front, reduceState is currently used as a non-persistent stream in some places (which is very good!). Still, it is also used for the “legacy stream”, where it acts the same as persistentReducedStream, but without the benefits of auto-subscribing and .state.


src/reducedStream.ts
```diff
@@ -0,0 +1,61 @@
...
+  reducers: RegisteredReducer<State, any>[],
+  { errorSubject, namespace, action$: a$ }: ReducedStreamOptions = {}
+): Observable<State> => {
+  const action$ = a$ ?? defaultAction$;
+  const filteredAction$ =
+    namespace === undefined ? action$ : action$.pipe(withNamespace(namespace));
+
+  const source$ = filteredAction$.pipe(
+    combineReducers(initialState, reducers, {
+      errorSubject: errorSubject,
```
holographic-principle, line 49: Maybe errorSubject should also be optional?

rafaa2: It is optional for user (when creating a stream but internally it is "using the default one")

src/reducedStream.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param name The name of the stream, used for placing a spy tag
+ * @param initialState The initial state of the stream
+ * @param reducers The reducers that build up the stream state
+ */
+export const reducedStream = <State>(
+  name: string,
+  initialState: State,
+  reducers: RegisteredReducer<State, any>[],
+  { errorSubject, namespace, action$: a$ }: ReducedStreamOptions = {}
```
chriskr, line 41: Just set the default value in the signature?
```TypeScript
{ errorSubject, namespace, action$ = defaultAction$ }: ReducedStreamOptions = {}
```
?

src/reducedStream.ts
```diff
@@ -0,0 +1,61 @@
...
+    combineReducers(initialState, reducers, {
+      errorSubject: errorSubject,
+      namespace: namespace,
+    }),
+    startWith(initialState),
+    shareReplay({
+      refCount: true,
+      bufferSize: 1,
+    }),
+    tag(name)
```
chriskr, line 57: Are we still using tagging? I thought we removed it?

src/reducedStream.ts
```diff
@@ -0,0 +1,61 @@
...
+ * It is guaranteed that the state stream will always emit a value upon subscription.
+ *
+ * The values emitted from the stream are shared between the subscribers,
+ * and the reducers are only ran once per input action.
+ *
+ * @param name The name of the stream, used for placing a spy tag
+ * @param initialState The initial state of the stream
+ * @param reducers The reducers that build up the stream state
+ */
+export const reducedStream = <State>(
```
chriskr, line 37: You should use this one now for `persistentReducedStream`, half of the body is the same.

src/reducedStream.ts
rafaa2: we are planning on Investigating it. but not deleted yet

src/reducedStream.ts
rafaa2: I think, if you don't provide the last argument, it will be by default = {}
so action$ will be undefined

rafaa2: Okay! Looks neater! I will change it

chriskr (approved):
Not tested, but LGTM.

src/internal/_reducedStream.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Creates a simple reduced state stream (without tagging, making sure to have initial state, nor replay)
+ *
+ * @param initialState The initial state of the stream
+ * @param reducers The reducers that build up the stream state
+ * @param {ReducedStreamOptions} options
+ * @returns
+ */
+export const _reducedStream = <State>(
```
holographic-principle, line 21: Is this a new naming convention?

I think the fact that it's in `/internal/` should be enough.

holographic-principle: I would personally name this file and function both `reducedStream`, then rename the import (to e.g. `reducedStreamInternal` if it clashes.

In the far past, we have discussed naming conventions for private variables and concluded that prefixing things with `_` is an anti-pattern in modern JavaScript.

@chriskr I hope you remember the same thing I do :)

michaltaberski (approved):

src/internal/_reducedStream.ts
rafaa2: @holographic-principle  Changed 👌🏼

PR #576, opened 2023-01-18
Properly type `useStream` when `defaultValue` provided

jsloat:
It currently shows the value of `useStream` as `T | NOT_YET_EMITTED` even when `defaultValue` is provided, which is wrong -- if `defaultValue` is provided, the return type will always be `T`.

Typing works as expected: 

**Without default value**
Screenshot 2023-01-18 at 15 56 29

**With default value**
Screenshot 2023-01-18 at 15 57 37

I've never worked on this repo before so I may need some help deploying, if this makes sense to you guys.


rafaa2 (approved):
Looks good

PR #599, opened 2023-03-10
fix: Observed type for consistency with TS

sseppola:


AlicjaMajewska (approved):
Beautiful 👌 
I would add the note the description like `Same like Awaited infers the type of value wrapped in a promise, Observed infers the type of value in Observable` to make the connection between Awaited and Observed more visible, it might help memorising/connecting dots

PR #605, opened 2023-03-20
feat: Refactor ActionWithPayload/ActionWithoutPayload to be uniform

sseppola:


src/routines.spec.ts
```diff
@@ -18,55 +18,67 @@ beforeAll(() => {
...
+// Add number representations to marble map
+for (let i = 0; i < 10; i++) {
+  // Create marble representations of the 10 first characters
+  // eg. { A: 'A', B: 'B', ...etc }
+  const charCodeA = 'A'.charCodeAt(0);
+  const letter = String.fromCharCode(charCodeA + i);
+  marbleMap[letter] = letter;
+  // Create marble representations of the 10 first numbers from 0
+  marbleMap[`${i}`] = i;
+}
```
sseppola, line 43: I had to change these tests because it was using the lengths of the action types.. which obviously changes when you enforce the `ActionName` type.
Now the tests are more robust.

src/types/helpers.tspec.ts
```diff
@@ -1,8 +1,34 @@
...
+type VoidFn = () => void;
+
 type ExtractPayload_extracts_payload = AssertTrue<
   IsExact<ExtractPayload<ActionCreatorWithPayload<Payload>>, Payload>
 >;
+type ExtractPayload_extracts_payload2 = AssertTrue<
+  IsExact<ExtractPayload<ActionCreatorWithPayload>, void>
+>;
+// type aaa = ExtractPayload<VoidFn>; // weakness of ExtractPayload
+// type aaa3 = ExtractPayload<unknown>; // weakness of ExtractPayload
```
sseppola, line 15: InferPayloadFromActionCreator is the same as ExtractPayload, although a bit more flexible.
Thinking of renaming the former to the latter, so we use the most flexible version.

rafaa2:
I don't think this is working correctly.
image

src/actionCreator.ts
```diff
@@ -1,15 +1,17 @@
...
- * recognize the generic, typed overload of this function.
+ * @param type A name for debugging purposes
+ * @template `Payload` - The payload type for the action
+ * @returns An action creator function that accepts a payload as input, and
+ *          returns a complete action object with that payload and a type unique
+ *          to this action creator
  */
-export const actionCreator: ActionCreatorFunc = (type: string) => {
-  const actionCreatorFn = (payload?: any) =>
+export const actionCreator = <Payload = void>(type: ActionName) => {
```
rafaa2, line 13:
```suggestion
export const actionCreator = <Payload = void>(
  type: ActionName
): ActionCreator<Payload> => {
```

sseppola:
@rafaa2 what's wrong with your screenshot? The actionCreator functions have a type attribute, and rxbeach relies on it.

rafaa2:
@sseppola But it should show ActionCreator<Payload> not the whole object

sseppola:
The type that appears is a function signature, so it is an action creator.
I don't know if I understand what you mean.

rafaa2:
@sseppola I am not talking about `type`

when you use the action creator, the new object should be displayed when you hover on it as ActionCreator instead of showing its  shape:

This:
image

instead of:
image

AlicjaMajewska (approved):
Great work 💯

PR #674, opened 2023-09-13
ARD-18156 add type mapping for one more param for derived stream

AlicjaMajewska:
We need one more param in ardoq front 🥳


chriskr (approved):
RxJs is now using dictionaries instead of arrays, which is IMO beter, we should follow that pattern in rxbeach, I think.

sseppola (approved):
The hoops we jump through to make TS happy 😂

PR #675, opened 2023-10-10
chore!: migrate package to ESM

oyvinmar:
## What

This change transitions the package to be distributed as an ESM (ECMAScript Module) instead of CommonJS. The main change is the output of ESM-compatible code by the TypeScript compiler and the modification of the module type in `package.json`.

Given that most bundlers now support ESM-distributed packages, I decided to forgo adding backward compatibility for CommonJS.

BREAKING CHANGE: The package is now distributed as an ESM module

## Why

The combination of CommonJS and Vite caused problems when introducing rxbeach in surveys (see: https://github.com/ardoq/ardoq-packages/pull/6393). Changing to ESM seems to have fixed the problem. 

## Test
I've tested the changes in both `ardoq-surveys` and  `ardoq-front` with the help of https://github.com/mweststrate/relative-deps. Will deploy and test in the test environment once released.


sseppola (approved):
As long as you've tested this manually it LGTM 👍 
Make sure you run the test suite as well, I had to do some work to get rxbeach to work smoothly with jest at some point.

OyvindSabo (approved):
Tested in `ardoq-surveys` and `ardoq-front`:nerd_face:

chriskr (approved):
Not tested anything, but LGTM.

